PR #6, opened 2016-08-06
Convert environment variables to string

### LINKED ISSUE #4: Flattened environment should always be a string
Currently it can be string, bool, int, or double.

catto:
This PR fixes #4.
Environment variables from a `matrix` will be also converted to string.


stjohnjohnson:
Excellent!!  I love that `JSON.stringify` doesn't actually change the contents, but simply casts it to a string.

👍 excellent

d2lam:
👍  LGTM

PR #7, opened 2016-08-08
Switching to a new screwdriver.yaml and validator format

stjohnjohnson:
This is because JSON and YAML do not guarantee object ordering

(also, we don't use the CLI anymore)

YAML keys are unordered in maps: http://yaml.org/spec/1.0/#id2558578
Recommendation for ordering keys in YAML: http://yaml.org/spec/1.0/#id2491203
Golang currently randomizes map key ordering: https://blog.golang.org/go-maps-in-action#TOC_7.


lib/phase/flatten.js
```diff
@@ -67,6 +67,52 @@ function cleanComplexEnvironment(jobs) {
...
+ *  - { name: "name", command: "command" }
+ * @method convertSteps
+ * @param  {Object}       jobs   List of jobs (name => job)
+ * @return {Object}              Updated list of jobs after up-converting
+ */
+function convertSteps(jobs) {
+    const newJobs = jobs;
+
+    Object.keys(newJobs).forEach((jobName) => {
+        const steps = clone(Hoek.reach(jobs, `${jobName}.steps`, {
```
nkatzman: should you be using `newJobs` here as well?

nkatzman: Or in line 85 should you be using `jobs` ?

nkatzman:
Seems reasonable 👍

test/index.test.js
```diff
@@ -100,7 +100,7 @@ describe('config parser', () => {
         it('returns an error if not enough steps', (done) => {
             parser(loadData('not-enough-commands.yaml'), (err) => {
                 assert.isNotNull(err);
-                assert.match(err.toString(), /"steps" requires at least one step/);
+                assert.match(err.toString(), /"steps" must contain at least 1 items/);
```
d2lam, line 103: `item`

d2lam:
Just small comment. LGTM 👍

jer:
👍

lib/phase/flatten.js
```diff
@@ -67,6 +67,52 @@ function cleanComplexEnvironment(jobs) {
...
+            default: []
+        }));
+
+        newJobs[jobName].commands = steps.map((step, index) => {
+            let name;
+            let command;
+
+            switch (typeof step) {
+            case 'object':
+                name = Object.keys(step).pop();
```
d2lam, line 96: Actually wouldn't this give you 'command'?

nkatzman, line 96: there is only 1 key in the object

d2lam, line 96: Shouldn't it be as simple as

```
name = step['name'],
command = step['command']
```

d2lam, line 96: @nkatzman  Yeah nvm I got mixed up between what the user provides and what the launcher expects

nkatzman, line 96: The key value that is here is a user input so you don't know what it is.

This code is trying to translate it into the `{ name: 'foo', command: 'bar' }` schema

PR #16, opened 2016-10-12
Don't fail when yaml is invalid

petey:
Converts config-parser to promise interface.
When a parse error occurs, the parser will return a valid config that spits out the error message, and exits with a failure.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,12 +2,17 @@
 
 const Yaml = require('js-yaml');
 const Hoek = require('hoek');
-const Async = require('async');
 
 const phaseValidateStructure = require('./lib/phase/structural');
 const phaseFlatten = require('./lib/phase/flatten');
 const phaseValidateFunctionality = require('./lib/phase/functional');
 const phaseGeneratePermutations = require('./lib/phase/permutation');
+
+const parseYaml = yaml => (new Promise((resolve) => {
+    const parsedYaml = Yaml.safeLoad(yaml);
+
+    resolve(parsedYaml);
+}));
 
 /**
  * Parse the configuration from a screwdriver.yaml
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -2,45 +2,50 @@
...
+        .then(doc => ({
             jobs: Hoek.reach(doc, 'jobs'),
             workflow: Hoek.reach(doc, 'workflow')
-        })
-    ], callback);
+        }))
+        .catch(err => ({
+            jobs: {
+                main: {
+                    image: 'node:4',
```
d2lam: Why `node:4`?

index.js
```diff
@@ -2,45 +2,50 @@
...
+                    steps: [{
```
d2lam: shouldn't this be `commands`

d2lam (changes requested):

index.js
petey: I dunno. I'll change it to `alpine`

index.js
petey: right you are then.

index.js
```diff
@@ -2,45 +2,52 @@
...
 const Yaml = require('js-yaml');
 const Hoek = require('hoek');
-const Async = require('async');
 
 const phaseValidateStructure = require('./lib/phase/structural');
 const phaseFlatten = require('./lib/phase/flatten');
 const phaseValidateFunctionality = require('./lib/phase/functional');
 const phaseGeneratePermutations = require('./lib/phase/permutation');
 
+const parseYaml = yaml => (new Promise((resolve) => {
```
stjohnjohnson: Can you make this one-line?

minzcmu (approved):
👍 
fellcat

stjohnjohnson (approved):
Looking good! 👍

d2lam (approved):

FenrirUnbound (approved):
👍 

Looks good to me

345t9873154

index.js
petey: fixed.

PR #17, opened 2016-10-13
On failure, main needs to be an array of permutations

petey:
Breaks validation if main is not an array


stjohnjohnson (approved):
Whoops.  👍

PR #19, opened 2016-10-23
Alpine linux doesn't work for us right now as our launcher cannot run in it

stjohnjohnson:
[buildpack-deps](https://hub.docker.com/r/_/buildpack-deps/) is the common base on a lot of core images (node, ruby, etc).

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ module.exports = function configParser(yaml) {
         .catch(err => ({
             jobs: {
                 main: [{
-                    image: 'alpine',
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
👍 

Seems fine to me

a purr-gato-ry ii - imgur

FenrirUnbound (approved):

petey:
👍  We will still probably need to figure out how to get launcher to work on alpine, though.

maybe_i_can_get_something_out_of_his_bad_playing

PR #30, opened 2017-02-25
fix: remove stuff for CLI

okuryu:
The CLI feature seems to removed in #7. These should no longer be necessary.


stjohnjohnson (approved):
👍 Thanks for cleaning it up!

PR #32, opened 2017-03-10
feat(470): Flatten templates

d2lam:
This merges the template into the job config so that user can use it. Individual job settings take precedence over templates settings. 

~This new version takes in an optional API URL to look up the templates specified in their `screwdriver.yaml`. If they don't specify the API URL and they don't use template in their `screwdriver.yaml`, it still works like before. Debating whether to make this a `major` or `minor` change.~

~Blocked by: this depends on the API endpoint to get template by `name` and `version`. [Third task in the TODO here.](https://github.com/screwdriver-cd/screwdriver/issues/470#issuecomment-285238348)~

Taking in templateFactory to retrieve template instead of making API call. 
Blocked by: 
    - model having the `getTemplate` function
    - data schema: https://github.com/screwdriver-cd/data-schema/pull/115
Related: https://github.com/screwdriver-cd/screwdriver/issues/470

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,23 +12,24 @@ const phaseGeneratePermutations = require('./lib/phase/permutation');
  * Parses a yaml file
  * @method parseYaml
  * @param  {String}  yaml Raw yaml
- * @return {Promise}      resoves POJO containing yaml data
+ * @return {Promise}      Resolves POJO containing yaml data
  */
 const parseYaml = yaml => (new Promise(resolve => resolve(Yaml.safeLoad(yaml))));
 
 /**
  * Parse the configuration from a screwdriver.yaml
  * @method configParser
- * @param  {String}   yaml      Contents of screwdriver.yaml
- * @param  {Function} callback  Function to call when done (error, { workflow, jobs })
+ * @param  {String}           yaml              Contents of screwdriver.yaml
+ * @param  {TemplateFactory}  templateFactory   Template Factory to get templates
+ * @returns {Promise}
  */
-module.exports = function configParser(yaml) {
+module.exports = function configParser(yaml, templateFactory) {
     // Convert from YAML to JSON
     return parseYaml(yaml)
         // Basic validation
         .then(phaseValidateStructure)
         // Flatten structures
-        .then(phaseFlatten)
+        .then(parsedDoc => phaseFlatten(parsedDoc, templateFactory))
         // Functionality validation
         .then(phaseValidateFunctionality)
         // Generate Permutations
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -2,14 +2,41 @@
 
 const Hoek = require('hoek');
 const clone = require('clone');
+const TEMPLATE_REGEX = require('screwdriver-data-schema').config.regex.FULL_TEMPLATE_NAME;
+const NAME_MATCH = 1;
+const VERSION_MATCH = 3;
+const LABEL_MATCH = 7;
 
 /**
- * Overlays the stuff specified in a specific job on top of the defaults in shared
+ * Merge oldJob into newJob
+ * "oldJob" takes precedence over "newJob". For ex: individual job settings vs shared settings
+ * @param  {Object}   newJob    Job to be merged into. For ex: shared settings
+ * @param  {Object}   oldJob    Job to merge. For ex: individual job settings
+ */
+function merge(newJob, oldJob) {
+    // Intialize new job with default fields (environment, settings, and secrets)
+    newJob.environment = newJob.environment || {};
+    newJob.settings = newJob.settings || {};
+
+    // Merge
+    Object.assign(newJob.environment, oldJob.environment || {});
+    Object.assign(newJob.settings, oldJob.settings || {});
+    newJob.image = oldJob.image || newJob.image;
+
+    // Merge secrets
+    const newSecrets = newJob.secrets || [];
+    const oldSecrets = oldJob.secrets || [];
+
+    newJob.secrets = [...new Set([...newSecrets, ...oldSecrets])];  // remove duplicate
+}
+
+/**
+ * Overlays the fields specified in a specific job on top of the defaults in shared
  *
  * @method flattenSharedIntoJobs
  * @param  {Job}          shared A kind of default Job template
- * @param  {Object}       jobs   List of jobs (name => job)
- * @return {Object}              Updated list of jobs after merging
+ * @param  {Object}       jobs   Object with all the jobs
+ * @return {Object}              New object with jobs after merging
  */
 function flattenSharedIntoJobs(shared, jobs) {
     const newJobs = {};
@@ -19,31 +46,83 @@ function flattenSharedIntoJobs(shared, jobs) {
         const oldJob = clone(jobs[jobName]);
 
         // Replace
-        ['image', 'matrix', 'steps'].forEach((key) => {
+        ['image', 'matrix', 'steps', 'template'].forEach((key) => {
             if (oldJob[key]) {
                 newJob[key] = oldJob[key];
             }
         });
 
-        if (!newJob.environment) {
-            newJob.environment = {};
-        }
-        if (!newJob.secrets) {
-            newJob.secrets = [];
-        }
-        if (!newJob.settings) {
-            newJob.settings = {};
-        }
-
-        // Merge
-        Object.assign(newJob.environment, oldJob.environment || {});
-        Object.assign(newJob.settings, oldJob.settings || {});
-        newJob.secrets = newJob.secrets.concat(oldJob.secrets || []);
-
+        merge(newJob, oldJob);
         newJobs[jobName] = newJob;
     });
 
     return newJobs;
+}
+
+/**
+ * Retrieve template and merge into job config
+ *
+ * @method mergeTemplateIntoJob
+ * @param  {String}           jobName           Job name
+ * @param  {Object}           jobConfig         Job config
+ * @param  {Object}           newJobs           Object with all the jobs
+ * @param  {TemplateFactory}  templateFactory   Template Factory to get templates
+ * @return {Promise}
+ */
+function mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory) {
+    const matched = TEMPLATE_REGEX.exec(jobConfig.template);
+    const name = matched[NAME_MATCH];
+    const version = matched[VERSION_MATCH];
+    const label = matched[LABEL_MATCH] ? matched[LABEL_MATCH].slice(1) : '';
+    const oldJob = jobConfig;
+
+    // Try to get the template
+    return templateFactory.getTemplate({ name, version, label })
+        .then((template) => {
+            if (!template) {
+                const labelText = label ? ` with label '${label}'` : '';
+
+                throw new Error(`Template ${name}@${version}${labelText} does not exist`);
+            }
+
+            const newJob = template.config;
+
+            merge(newJob, oldJob);
+            delete newJob.template;
+            newJobs[jobName] = newJob;
+
+            return null;
+        });
+}
+
+/**
+ * Goes through each job and if template is specified, then merge into job config
+ *
+ * @method flattenTemplates
+ * @param  {Object}           jobs              Object with all the jobs
+ * @param  {TemplateFactory}  templateFactory   Template Factory to get templates
+ * @return {Promise}          Resolves to new object with jobs after merging templates
+ */
+function flattenTemplates(jobs, templateFactory) {
+    const newJobs = {};
+    const templates = [];
+
+    // eslint-disable-next-line
+    Object.keys(jobs).forEach((jobName) => {
+        const jobConfig = clone(jobs[jobName]);
+        const templateConfig = jobConfig.template;
+
+        // If template is specified, then merge
+        if (templateConfig) {
+            templates.push(mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory));
+        } else {
+            newJobs[jobName] = jobConfig;   // Otherwise just use jobConfig
+        }
+    });
+
+    // Wait until all promises are resolved
+    return Promise.all(templates)
+        .then(() => newJobs);
 }
 
 /**
@@ -52,8 +131,8 @@ function flattenSharedIntoJobs(shared, jobs) {
  * This is because YAML allows you to define complex structures easily, but our input to the shell
  * is just simple strings (environment variables).
  * @method cleanComplexEnvironment
- * @param  {Object}       jobs   List of jobs (name => job)
- * @return {Object}              Updated list of jobs after cleaning
+ * @param  {Object}       jobs   Object with all the jobs
+ * @return {Object}              Updated object with jobs after cleaning
  */
 function cleanComplexEnvironment(jobs) {
     Object.keys(jobs).forEach((jobName) => {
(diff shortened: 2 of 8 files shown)


lib/phase/flatten.js
```diff
@@ -19,31 +43,90 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+function mergeTemplateIntoJob(jobName, jobConfig, newJobs, apiUrl) {
+    const matched = TEMPLATE_REGEX.exec(jobConfig.template);
+    const name = matched[NAME_MATCH];
+    const version = matched[VERSION_MATCH];
+    const breaker = new Fusebox(request, {});
+    const oldJob = jobConfig;
+
+    // Try to get the template
+    return breaker.runCommand({
+        url: `${apiUrl}/v4/templates?name=${name}&version=${version}`,
```
stjohnjohnson: Why are we using query parameters here?  They cannot be cached.  Shouldn't this be `/templates/$name/$version/`?

Also, shouldn't we be escaping them?

lib/phase/flatten.js
```diff
@@ -19,31 +43,90 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+        method: 'GET',
+        json: true
+    }).then((response) => {
+        if (response.statusCode === 200) {
+            const templates = response.body;
+
+            if (templates.length <= 0) {
```
stjohnjohnson: Why does this return an array?

index.js
```diff
@@ -12,23 +12,24 @@ const phaseGeneratePermutations = require('./lib/phase/permutation');
...
+ * @return {Promise}      Resolves POJO containing yaml data
  */
 const parseYaml = yaml => (new Promise(resolve => resolve(Yaml.safeLoad(yaml))));
 
 /**
  * Parse the configuration from a screwdriver.yaml
  * @method configParser
  * @param  {String}   yaml      Contents of screwdriver.yaml
- * @param  {Function} callback  Function to call when done (error, { workflow, jobs })
+ * @param  {String}   apiUrl    API Url to retrieve templates (default to https://api.screwdriver.cd)
```
stjohnjohnson: Why are we now making API calls from within the config parser?

stjohnjohnson (changes requested):

index.js
d2lam: How else can we look up templates? Intially I was thinking passing in a templateFactory, but that would break the ability to run this as command line. Discussed with @petey and decided to go with making API calls in here. What other options are you thinking?

lib/phase/flatten.js
d2lam: Haven't figured out the endpoint yet. This is just a placeholder

index.js
stjohnjohnson: This doesn't run as a command line.  We removed all the code for that.

petey: Sorry, didn't know that.

package.json
```diff
@@ -43,16 +43,19 @@
...
     "eslint": "^3.2.2",
     "eslint-config-screwdriver": "^2.0.0",
-    "eslint-plugin-import": "^2.0.0",
-    "jenkins-mocha": "^4.0.0"
+    "jenkins-mocha": "^4.0.0",
+    "sinon": "^1.17.7",
+    "sinon-as-promised": "^4.0.2"
   },
   "dependencies": {
+    "circuit-fuses": "^2.2.1",
```
tkyi: can remove this package since we're not making calls to the API anymore

package.json
```diff
@@ -43,16 +43,19 @@
...
     "clone": "^2.0.0",
     "hoek": "^4.0.1",
     "joi": "^10.0.5",
     "js-yaml": "^3.6.1",
     "keymbinatorial": "^1.0.0",
-    "screwdriver-data-schema": "^16.0.1",
+    "request": "^2.80.0",
```
tkyi: can also remove this package since we're not making calls to the API anymore

index.js
```diff
@@ -12,23 +12,24 @@ const phaseGeneratePermutations = require('./lib/phase/permutation');
...
  */
-module.exports = function configParser(yaml) {
+module.exports = function configParser(yaml, templateFactory) {
     // Convert from YAML to JSON
     return parseYaml(yaml)
         // Basic validation
         .then(phaseValidateStructure)
         // Flatten structures
-        .then(phaseFlatten)
+        .then(docs => phaseFlatten(docs, templateFactory))
```
FenrirUnbound: `parsedDoc` instead of `docs` so it matches the call signature

lib/phase/flatten.js
```diff
@@ -2,9 +2,32 @@
...
 /**
- * Overlays the stuff specified in a specific job on top of the defaults in shared
+ * Merge oldJob into newJob
+ * "oldJob" takes precedence over "newJob". For ex: individual job settings vs shared settings
+ * @param  {Object}   newJob    Job to be merged into. For ex: shared settings
+ * @param  {Object}   oldJob    Job to merge. For ex: individual job settings
+ */
+function merge(newJob, oldJob) {
+    // Intialize new job with default fields
+    newJob.environment = newJob.environment || {};
```
FenrirUnbound: Would be nice to document these fields

lib/phase/flatten.js
```diff
@@ -2,9 +2,32 @@
...
+    newJob.secrets = newJob.secrets || [];
+    newJob.settings = newJob.settings || {};
+
+    // Merge
+    Object.assign(newJob.environment, oldJob.environment || {});
+    Object.assign(newJob.settings, oldJob.settings || {});
+    newJob.secrets = newJob.secrets.concat(oldJob.secrets || []);
```
FenrirUnbound: Is there something that removes duplicates?

lib/phase/flatten.js
```diff
@@ -2,9 +2,32 @@
...
+ */
+function merge(newJob, oldJob) {
+    // Intialize new job with default fields
+    newJob.environment = newJob.environment || {};
+    newJob.secrets = newJob.secrets || [];
+    newJob.settings = newJob.settings || {};
+
+    // Merge
+    Object.assign(newJob.environment, oldJob.environment || {});
+    Object.assign(newJob.settings, oldJob.settings || {});
```
FenrirUnbound, line 23: The default value for `.secrets` is an array: https://github.com/screwdriver-cd/config-parser/pull/32/files#diff-daa88a14d1ada8a3aa7d881f9424ba2cR19 ~~Is this also supposed to be an array?~~

EDIT: Looking at the old code, it's supposed to be an array: https://github.com/screwdriver-cd/config-parser/pull/32/files#diff-daa88a14d1ada8a3aa7d881f9424ba2cL41

EDIT2: nevermind, the diff I was reading was wrong to begin with. This is fine

FenrirUnbound: Why are there no tests for the alternative values?

lib/phase/flatten.js
d2lam: Yeah we should remove duplicates. Currently, it doesn't when we merge `shared` and individual job:
screen shot 2017-03-14 at 2 23 30 pm

I should fix that.

lib/phase/flatten.js
```diff
@@ -19,41 +42,91 @@ function flattenSharedIntoJobs(shared, jobs) {
...
     return newJobs;
 }
 
 /**
+ * Retrieve template and merge into job config
+ *
+ * @method mergeTemplateIntoJob
+ * @param  {String}           jobName           Job name
+ * @param  {Object}           jobConfig         Job config
+ * @param  {Object}           newJobs           List of new jobs
```
FenrirUnbound: > List of new jobs

Is this an Array or an Object?

lib/phase/flatten.js
```diff
@@ -19,41 +42,91 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+ * @param  {TemplateFactory}  templateFactory   Template Factory to get templates
+ * @return {Promise}
```
FenrirUnbound: What does the Promise resolve to?

lib/phase/flatten.js
```diff
@@ -19,41 +42,91 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+    const templates = [];
+
+    // eslint-disable-next-line
+    Object.keys(jobs).forEach((jobName) => {
+        const jobConfig = clone(jobs[jobName]);
+        const templateConfig = jobConfig.template;
+
+        // If template is specified, then merge
+        if (templateConfig) {
+            templates.push(mergeTemplateIntoJob(jobName, jobConfig, newJobs, apiUrl));
```
FenrirUnbound: The method signature is wrong: `mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory)`

lib/phase/flatten.js
d2lam: What do you mean by alternative values?

lib/phase/flatten.js
```diff
@@ -87,7 +160,7 @@ function cleanComplexEnvironment(jobs) {
  *  - { name: "name", command: "command" }
  * @method convertSteps
  * @param  {Object}       jobs   List of jobs (name => job)
- * @return {Object}              Updated list of jobs after up-converting
+ * @return {Promise}             Updated list of jobs after up-converting
```
FenrirUnbound: I don't see anything in this function that returns a Promise.

lib/phase/flatten.js
```diff
@@ -19,41 +42,91 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+
+/**
  * Converts complex environment variables like objects or arrays into JSON-ified strings
  *
  * This is because YAML allows you to define complex structures easily, but our input to the shell
  * is just simple strings (environment variables).
  * @method cleanComplexEnvironment
  * @param  {Object}       jobs   List of jobs (name => job)
- * @return {Object}              Updated list of jobs after cleaning
+ * @return {Promise}             Updated list of jobs after cleaning
```
FenrirUnbound: I don't see anything here that returns a Promise.

lib/phase/flatten.js
d2lam: oops good catch. Forgot to change it.

lib/phase/flatten.js
FenrirUnbound: There are no tests that assert the default values; nothing asserts what happens with `oldJob.settings` is omitted.

FenrirUnbound (changes requested):
I keep forgetting to add the comments to a review 😞

lib/phase/flatten.js
d2lam: You're right. Actually it's still returning an object

lib/phase/flatten.js
d2lam: Objects with jobs as key. I'm using the same comments as here: https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/flatten.js#L11

I can make it more clear.

lib/phase/flatten.js
d2lam: This one covers it: https://github.com/screwdriver-cd/config-parser/blob/9bed04a9e87e33d1960abaaaf8445439a49a994d/test/index.test.js#L125-L127

No `settings` in individual job config: https://github.com/screwdriver-cd/config-parser/blob/9bed04a9e87e33d1960abaaaf8445439a49a994d/test/data/basic-job-with-template.yaml

But the merged one has it `settings: {}`: https://github.com/screwdriver-cd/config-parser/blob/9bed04a9e87e33d1960abaaaf8445439a49a994d/test/data/basic-job-with-template.json

FenrirUnbound: Gotcha. The test names weren't really descriptive

lib/phase/flatten.js
```diff
@@ -2,14 +2,41 @@
...
+
+/**
+ * Overlays the fields specified in a specific job on top of the defaults in shared
  *
  * @method flattenSharedIntoJobs
  * @param  {Job}          shared A kind of default Job template
- * @param  {Object}       jobs   List of jobs (name => job)
- * @return {Object}              Updated list of jobs after merging
+ * @param  {Object}       jobs   Object with all the jobs
+ * @return {Object}              Updated object with jobs after merging
```
FenrirUnbound: Does it modify the `jobs` object? It would be more clear if it described that it returns a new object and not a modified `jobs` object.

FenrirUnbound (approved):
👍 

Looks good as is

glulf0leoely

lib/phase/flatten.js
d2lam: Ok updated comment

lib/phase/flatten.js
```diff
@@ -2,14 +2,41 @@
...
+    // Merge
+    Object.assign(newJob.environment, oldJob.environment || {});
+    Object.assign(newJob.settings, oldJob.settings || {});
+    newJob.image = oldJob.image || newJob.image;
+
+    // Merge secrets
+    const newSecrets = newJob.secrets || [];
+    const oldSecrets = oldJob.secrets || [];
+
+    newJob.secrets = [...new Set([...newSecrets, ...oldSecrets])];  // remove duplicate
```
petey, line 30: Interesting, I had to look this one up.

lib/phase/flatten.js
```diff
@@ -2,14 +2,41 @@
...
+const LABEL_MATCH = 7;
 
 /**
- * Overlays the stuff specified in a specific job on top of the defaults in shared
+ * Merge oldJob into newJob
+ * "oldJob" takes precedence over "newJob". For ex: individual job settings vs shared settings
+ * @param  {Object}   newJob    Job to be merged into. For ex: shared settings
+ * @param  {Object}   oldJob    Job to merge. For ex: individual job settings
+ */
+function merge(newJob, oldJob) {
```
petey, line 16: Is there a reason we do not use something like `hoek.merge`?

lib/phase/flatten.js
```diff
@@ -19,41 +46,93 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+ * @method flattenTemplates
+ * @param  {Object}           jobs              Object with all the jobs
+ * @param  {TemplateFactory}  templateFactory   Template Factory to get templates
+ * @return {Promise}          Resolves to new object with jobs after merging templates
+ */
+function flattenTemplates(jobs, templateFactory) {
+    const newJobs = {};
+    const templates = [];
+
+    // eslint-disable-next-line
```
petey, line 110: why is this comment here?

petey (approved):

test/data/basic-job-with-template.yaml
```diff
@@ -0,0 +1,7 @@
+jobs:
+    main:
+        template: mytemplate@1.2.3
```
minzcmu, line 3: Can we add some steps in the job to verify that the template steps override the job step?

minzcmu (changes requested):

test/data/basic-job-with-template.yaml
d2lam: Let's not handle overriding yet. For overriding, we need to go through each step in individual job and look up the exact step inside template, then override that specific step while keeping everything else the same. Since this logic is more complicated than replacing the whole step object, we will do it in 2nd pass when we handle override & wrapping of steps.

lib/phase/flatten.js
d2lam: This function does more than just merging, it also initialize fields it does not have, and remove duplicates if merged.  This is the console log from some tests. 

Using this function:
```
{
  "environment": {
    "NODE_ENV": "test"
  },
  "image": "node:{{NODE_VERSION}}",
  "matrix": {
    "NODE_VERSION": [
      4,
      5,
      6
    ]
  },
  "steps": [
    {
      "init": "npm install"
    },
    {
      "test": "npm test"
    }
  ],
  "settings": {},
  "secrets": []
}
```

Using hoek.merge
```
{
  "environment": {
    "NODE_ENV": "test"
  },
  "image": "node:{{NODE_VERSION}}",
  "matrix": {
    "NODE_VERSION": [
      4,
      5,
      6,
      4,
      5,
      6
    ]
  },
  "steps": [
    {
      "init": "npm install"
    },
    {
      "test": "npm test"
    },
    {
      "init": "npm install"
    },
    {
      "test": "npm test"
    }
  ]
}

```

lib/phase/flatten.js
d2lam: oops leftover from something else

minzcmu (approved):

PR #33, opened 2017-03-30
fix: output an array of error messages when parse fails

petey:
While the api documentation provides for an errors array, the config parser hasn't ever populated it. This PR adds an error to that array.


tkyi (approved):
Ooh

d2lam (approved):

PR #36, opened 2017-06-12
fix(599): Throw error when screwdriver.yaml does not exist

tkyi:
Add logic to check if screwdriver.yaml exists. If it doesn't, throw an error.

Related to https://github.com/screwdriver-cd/screwdriver/issues/599
https://github.com/screwdriver-cd/screwdriver/issues/593


minzcmu (approved):
👍 LGTM but probably won't fixed https://github.com/screwdriver-cd/screwdriver/issues/599

d2lam (approved):

PR #37, opened 2017-06-14
feat(580): support for annotations

r3rastogi:
Read in the annotations as described in screwdriver-cd/data-schema#145.

Pipeline-level annotations are separate than job-level annotations. Job-level annotations override shared annotations.

Related to screwdriver-cd/screwdriver#580.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,10 +44,12 @@ module.exports = function configParser(yaml, templateFactory) {
         .then(phaseGeneratePermutations)
         // Output in the right format
         .then(doc => ({
+            annotations: Hoek.reach(doc, 'annotations', { default: {} }),
             jobs: Hoek.reach(doc, 'jobs'),
             workflow: Hoek.reach(doc, 'workflow')
         }))
         .catch(err => ({
+            annotations: {},
             jobs: {
                 main: [{
                     image: 'node:6',
--- a/test/data/shared-job-annotations.yaml
+++ b/test/data/shared-job-annotations.yaml
@@ -1,0 +1,12 @@
+shared:
+    annotations:
+        foo: a
+        bar: b
+
+jobs:
+    main:
+        annotations:
+            foo: c
+        image: node:4
+        steps:
+            - install: npm install
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -15,10 +15,12 @@ const LABEL_MATCH = 7;
  */
 function merge(newJob, oldJob) {
     // Intialize new job with default fields (environment, settings, and secrets)
+    newJob.annotations = newJob.annotations || {};
     newJob.environment = newJob.environment || {};
     newJob.settings = newJob.settings || {};
 
     // Merge
+    Object.assign(newJob.annotations, oldJob.annotations || {});
     Object.assign(newJob.environment, oldJob.environment || {});
     Object.assign(newJob.settings, oldJob.settings || {});
     newJob.image = oldJob.image || newJob.image;
--- a/lib/phase/permutation.js
+++ b/lib/phase/permutation.js
@@ -19,6 +19,9 @@ module.exports = validatedDoc => (
         // Generate permutations for each job
         Object.keys(doc.jobs).forEach((jobName) => {
             const job = {
+                annotations: Hoek.reach(doc, `jobs.${jobName}.annotations`, {
+                    default: {}
+                }),
                 image: Hoek.reach(doc, `jobs.${jobName}.image`),
                 commands: Hoek.reach(doc, `jobs.${jobName}.commands`),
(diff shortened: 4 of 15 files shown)


tkyi (commented):
What happens if a user specifies a pipeline level annotation that conflicts with a job level annotation? Or which one takes priority between pipeline level annotations and shared annotations?
Do we have test cases for those?

r3rastogi:
Pipeline-level annotations and job-level annotations are separate entities. Only shared/job will collapse into job-level (and job-level will override shared). I've included additional tests to further illustrate this point.

test/data/shared-job-annotations.yaml
```diff
@@ -0,0 +1,13 @@
+shared:
+    annotations:
+        foo: a
+        bar: b
+
+jobs:
+    main:
+        annotations:
+            foo: c
+            bar: d
```
d2lam: Can we modify this so that the job level has `foo`, and not `bar`? So that we know it will pick up `bar` from `shared`, and not completely override.

tkyi:
@r3rastogi Oh okay I see. So for our specific use case, we'd want to put the executor annotations at the job/shared level, never the pipeline level. But there will be other features in the future where it would make sense to put things in the pipeline level. Annotations that belong in pipeline vs shared/job level should never really overlap.

index.js
```diff
@@ -44,10 +44,12 @@ module.exports = function configParser(yaml, templateFactory) {
         .then(phaseGeneratePermutations)
         // Output in the right format
         .then(doc => ({
+            annotations: Hoek.reach(doc, 'annotations', { default: {} }),
```
minzcmu, line 47: For the pipeline level annotation, I don't think we can do it now without sequelize 4.x since we cannot create a new field `annotations`. :(

When it goes through the pipeline.sync, it will only update the workflow for the pipeline.
https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L251

It probably won't break anything but it's also not doing anything.. we won't be able to see it from `/pipelines/id`.

stjohnjohnson (approved):
Nice!

test/data/shared-job-annotations.yaml
r3rastogi: Sounds good. Updated in next push.

r3rastogi:
@tkyi yes, that's correct

index.js
r3rastogi: If we've already included it in `data-schema`, why does it also have to be in `models`?

d2lam: @r3rastogi The config-parser only parsers the yaml.

For jobs, it goes through the parser, then store the environment, settings, etc. under `permutations`. And because`permutations` is just an object, we can just add additional fields without adding a new column in the database. Example: https://api.screwdriver.cd/v4/jobs/1

For pipeline, we need to add a new field, because it's not under another object. Sorry I remember telling you we don't need to. 

Basically, you don't need to modify the jobs table, but you need to modify the pipeline table. And since sequelize 4.x doesn't work yet, we can't add a new field. I think it's best to not implement pipeline annotations for now. And when we get sequelize to work, then we can add it.

r3rastogi: Ahh got it. Thanks for the explanations @d2lam and @minz1027 .

I think we should leave this code as-is for now, and add in the `models` in a future PR to `data-schema` when relevant. As @minz1027 said, this won't break anything. I'll add a note to the original issue (https://github.com/screwdriver-cd/screwdriver/issues/580) indicating that we still need to modify the`pipeline` database.

d2lam: @r3rastogi  According to @FenrirUnbound, you should be able to manually alter the table to have the field. I'm not a fan of having code that is not doing anything. It's a little confusing. I think we should either alter the table, or remove code that is not doing anything. 

@FenrirUnbound  can you help reetika update the table? :)

FenrirUnbound: It's a SQL table, so just add a column with a sane default value

r3rastogi: @FenrirUnbound How do you get to the table? Are their onboarding or other docs for that?

r3rastogi: Adding a new column using the following SQL command:
`ALTER TABLE "public"."pipelines" ADD COLUMN "annotations" text;`

@FenrirUnbound annotations are an optional field, so I don't think having a default value is necessary.

r3rastogi: Pipelines table now looks like:
image

d2lam: Cool. I see it in production tables. I think you also need it for `beta`

d2lam (approved):
LGTM after the beta table is updated.

index.js
r3rastogi: Oops, didn't realize that was a thing. Added to `beta_pipelines`. I'm not sure what `beta_rc_pipelines` is though, so I didn't modify that one.

PR #38, opened 2017-06-23
feat(423): Allow parsing of description in config-parser

joelseq:
- Add basic job with description yaml file to test/data
- Allow parsing of description by modifying flatten and permutation
  phases

### DIFF
--- a/lib/phase/permutation.js
+++ b/lib/phase/permutation.js
@@ -22,16 +22,21 @@ module.exports = validatedDoc => (
                 annotations: Hoek.reach(doc, `jobs.${jobName}.annotations`, {
                     default: {}
                 }),
-                image: Hoek.reach(doc, `jobs.${jobName}.image`),
                 commands: Hoek.reach(doc, `jobs.${jobName}.commands`),
-                secrets: Hoek.reach(doc, `jobs.${jobName}.secrets`),
                 environment: Hoek.reach(doc, `jobs.${jobName}.environment`, {
                     default: {}
                 }),
+                image: Hoek.reach(doc, `jobs.${jobName}.image`),
+                secrets: Hoek.reach(doc, `jobs.${jobName}.secrets`),
                 settings: Hoek.reach(doc, `jobs.${jobName}.settings`, {
                     default: {}
                 })
             };
+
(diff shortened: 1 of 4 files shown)


lib/phase/permutation.js
```diff
@@ -22,16 +22,21 @@ module.exports = validatedDoc => (
...
                     default: {}
                 }),
+                image: Hoek.reach(doc, `jobs.${jobName}.image`),
+                secrets: Hoek.reach(doc, `jobs.${jobName}.secrets`),
                 settings: Hoek.reach(doc, `jobs.${jobName}.settings`, {
                     default: {}
                 })
             };
+
+            if ('description' in doc.jobs[jobName]) {
```
joelseq: This is the only way I could come up with for adding description without breaking / having to modify all the other tests. Is this okay?

test/index.test.js
```diff
@@ -271,6 +271,21 @@ describe('config parser', () => {
...
+            parser(loadData('basic-job-with-description.yaml'))
+                .then((data) => {
+                    assert.isDefined(data.jobs.main[0].description);
+                })
+        );
+
+        it('has the correct description', () =>
+            parser(loadData('basic-job-with-description.yaml'))
+                .then((data) => {
+                    assert.equal(data.jobs.main[0].description,
```
FenrirUnbound: You can combine this `assert.equal` in the previous `it` block

FenrirUnbound (changes requested):

lib/phase/permutation.js
d2lam: There is a difference between this and `if (doc.jobs[jobName].description)`
Does it make sense to set the description to `''`, or should we just not set it?

petey: This would be better.
```
description: Hoek.reach(doc, `jobs.${jobName}.description`, { default: '' })
```

joelseq: Yeah I guess it doesn't make sense if the description is an empty string or a falsy value. I just kept the option in case the description field was provided but I'll change it.

petey (changes requested):

petey (approved):

d2lam (approved):

PR #39, opened 2017-07-06
feat: Add support for wrapping steps in template

d2lam:
Allow wrapping around steps inside template. 
**Example:**
Inside template:
```yaml
   steps: 
       - test: echo test
       - publish: echo publish
```

Inside user's screwdriver.yaml:
```yaml
main:
   template: mytemplate@1
   steps: 
       - pretest: echo hello
       - posttest: echo goodbye
```

Translates into: 
```yaml
   steps:
       - pretest: echo hello
       - test: echo test
       - posttest: echo goodbye
       - publish: echo publish
```

Question: If the step `test` doesn't exist, should we error, or should we ignore it? With this implementation, it will ignore it. 

Related: https://github.com/screwdriver-cd/screwdriver/issues/616

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -10,10 +10,11 @@ const LABEL_MATCH = 7;
 /**
  * Merge oldJob into newJob
  * "oldJob" takes precedence over "newJob". For ex: individual job settings vs shared settings
- * @param  {Object}   newJob    Job to be merged into. For ex: shared settings
- * @param  {Object}   oldJob    Job to merge. For ex: individual job settings
+ * @param  {Object}   newJob        Job to be merged into. For ex: shared settings
+ * @param  {Object}   oldJob        Job to merge. For ex: individual job settings
+ * @param  {Boolean}  fromTemplate  Whether this is merged from template. If true, perform extra actions such as wrapping.
  */
-function merge(newJob, oldJob) {
+function merge(newJob, oldJob, fromTemplate) {
     // Intialize new job with default fields (environment, settings, and secrets)
     newJob.annotations = newJob.annotations || {};
     newJob.environment = newJob.environment || {};
@@ -30,6 +31,42 @@ function merge(newJob, oldJob) {
     const oldSecrets = oldJob.secrets || [];
 
     newJob.secrets = [...new Set([...newSecrets, ...oldSecrets])];  // remove duplicate
+
+    if (fromTemplate && oldJob.steps) {
+        let stepName;
+        let preStepName;
+        let postStepName;
+        const mergedSteps = [];
+
+        // convert steps from oldJob from array to object for faster lookup
+        const oldSteps = oldJob.steps.reduce((obj, item) => {
+            const key = Object.keys(item)[0];
+
+            obj[key] = item[key];
+
+            return obj;
+        }, {});
+
+        for (let i = 0; i < newJob.steps.length; i += 1) {
+            stepName = Object.keys(newJob.steps[i])[0];
+            preStepName = `pre${stepName}`;
+            postStepName = `post${stepName}`;
+
+            // add pre-step
+            if (oldSteps[preStepName]) {
+                mergedSteps.push({ [preStepName]: oldSteps[preStepName] });
+            }
+
+            mergedSteps.push(newJob.steps[i]);
+
+            // add post-step
+            if (oldSteps[postStepName]) {
+                mergedSteps.push({ [postStepName]: oldSteps[postStepName] });
+            }
+        }
+
+        newJob.steps = mergedSteps;
+    }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


tkyi:
I think ignoring it is fine.

lib/phase/flatten.js
```diff
@@ -13,7 +13,7 @@ const LABEL_MATCH = 7;
  * @param  {Object}   newJob    Job to be merged into. For ex: shared settings
  * @param  {Object}   oldJob    Job to merge. For ex: individual job settings
  */
-function merge(newJob, oldJob) {
+function merge(newJob, oldJob, fromTemplate) {
```
FenrirUnbound: Add `fromTemplate` to method documentation

FenrirUnbound: If it's a `Boolean`, I would strongly urge you to reconsider changing the method's parameters then

lib/phase/flatten.js
```diff
@@ -30,6 +30,42 @@ function merge(newJob, oldJob) {
...
+
+    if (fromTemplate && oldJob.steps) {
+        let stepName;
+        let preStepName;
+        let postStepName;
+        const mergedSteps = [];
+
+        // convert steps from oldJob from array to object for faster lookup
+        const oldSteps = oldJob.steps.reduce((obj, item) => {
+            const key = Object.keys(item)[0];
```
FenrirUnbound, line 43: Order isn't guaranteed

lib/phase/flatten.js
```diff
@@ -30,6 +30,42 @@ function merge(newJob, oldJob) {
...
+
+            obj[key] = item[key];
+
+            return obj;
+        }, {});
+
+        for (let i = 0; i < newJob.steps.length; i += 1) {
+            stepName = Object.keys(newJob.steps[i])[0];
```
FenrirUnbound, line 51: Again, order isn't guaranteed with `Object.keys()`

FenrirUnbound (changes requested):

lib/phase/flatten.js
ian-fox, line 43: This is the only bit I don't understand. What do the elements of `oldJob.steps` look like?

FenrirUnbound:
> Question: If the step `test` doesn't exist, should we error, or should we ignore it? With this implementation, it will ignore it.

The `config-parser` is pretty strict when you add in extra data, like adding a key to the `jobs` stanza that has no effect. That's the only reason why I'm leaning towards throwing an error. In the end, I don't really mind which implementation we end up going.

lib/phase/flatten.js
d2lam: I don't think it matters. There is only 1 key.

d2lam: It's a little confusing. oldJob.steps looks like this:
```
[ { pretest: 'echo pre-test' },
  { postinstall: 'echo post-install' } ]
```

lib/phase/flatten.js
d2lam: Same comment as above too. There is only 1 key in this object, and it's the `stepName`.

lib/phase/flatten.js
d2lam: Not sure what you mean? Like change the param to a JSON object with all these 3 things? Or you mean to change the name of this param?

lib/phase/flatten.js
ian-fox: Ah, okay. Always an object with one key?

Edit: didn't see the comment above. Makes sense in that case.

lib/phase/flatten.js
ian-fox: I think changing the param to be a `config` object or something, since a boolean as the last argument will default to `false` if the function gets called without it. Unless we want `fromTemplate=false` to be the default, I guess.

ian-fox: In either case though, should be in documentation block just above the function

lib/phase/flatten.js
d2lam: Yup. It's because when we write screwdriver.yaml it's:

```
steps:
     - test: echo test
     - publish: echo publish
```

ian-fox: Ahhhh okay that makes a lot of sense now! Thanks 😄

lib/phase/flatten.js
d2lam: Yeah I wanted it to be `false` by default since this is also called somewhere else, and I don't want to mess merging shared into jobs: https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/flatten.js#L57

lib/phase/flatten.js
FenrirUnbound: If you know the keyname, why not just use the keyname

FenrirUnbound: You might as well make it `newJob.steps[i].stepName`

FenrirUnbound: Nevermind. You *don't* know the keyname, just that it has one key. It's not glaringly obvious

d2lam: I don't have `stepName`. It is coming from `Object.keys(newJob.steps[i])[0]`

FenrirUnbound (approved):
👍 

c0fprj4 - imgur

minzcmu (approved):
👍 LGTM

PR #40, opened 2017-07-10
feat: support overriding steps for templates

d2lam:
Allows override steps for templates

```
main:
   template: mytemplate@1
   steps: 
       - publish: echo not publishing
```

Translate into:
```
   steps:
       - test: echo test
       - publish: echo not publishing
```

Related: https://github.com/screwdriver-cd/screwdriver/issues/616

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -57,7 +57,12 @@ function merge(newJob, oldJob, fromTemplate) {
                 mergedSteps.push({ [preStepName]: oldSteps[preStepName] });
             }
 
-            mergedSteps.push(newJob.steps[i]);
+            // override step
+            if (oldSteps[stepName]) {
+                mergedSteps.push({ [stepName]: oldSteps[stepName] });
+            } else {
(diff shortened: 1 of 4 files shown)


petey (approved):
`newJob` and `oldJob` are becoming somewhat confusing monikers for the inputs. The new/old variables make this additional snippet even more confusing without reading more context. We might want to refactor this in the future to something like `job` and `template` (or `jobTemplate`). Then you get something like `templateSteps` instead of `oldSteps` which makes a lot more sense with limited context.

d2lam:
@petey I agree with you. It's pretty confusing. This function is used by merging `shared` as well, which is why I couldn't rename it to something like `templateJob` or `templateSteps`. If you have some suggestions for the names, I would be happy to incorporate it.

tkyi:
`globalJobSettings` and `localJobSettings`? Or just `globalJob` and `localJob`? Arg idk.

minzcmu (approved):
👍 I think it's alright for now. Probably refactor the name later?

PR #43, opened 2017-08-19
feat: Allow Multiple-Documents

stjohnjohnson:
By default, load the first document.  Otherwise, if this is a multi-doc, look for a `version: 4` to know which document we want to load.

Depends on: https://github.com/screwdriver-cd/data-schema/pull/172

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ function parseYaml(yaml) {
         'create a screwdriver.yaml and try to rerun your build.');
     }
 
-    return new Promise(resolve => {
+    return new Promise((resolve) => {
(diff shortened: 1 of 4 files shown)


d2lam (approved):
LGTM 👍 
Let me try to plug this in locally first

tkyi (approved):
👍

d2lam:
Cool! Tested it out locally and looks like it's working. 
I tried multiple docs with no `version`:
screen shot 2017-08-18 at 5 35 20 pm

With `version: 4` it runs normal steps.  👍

PR #45, opened 2017-09-29
feat(723): supports new workflow design

d2lam:
- Support new workflow format
- If there is no `requires` specified, will assume it's an old version
- We should come back and get rid of some of the code once we switch completely to the new format

Blocked by: https://github.com/screwdriver-cd/data-schema/pull/182
Related: https://github.com/screwdriver-cd/screwdriver/issues/723

### DIFF
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -181,14 +181,17 @@ function validateWorkflow(doc) {
 module.exports = flattenedDoc => (
     new Promise((resolve, reject) => {
         const doc = flattenedDoc;
+        const hasRequires = Object.keys(doc.jobs).some(jobName => doc.jobs[jobName].requires);
         let errors = [];
 
         // Jobs
         errors = errors.concat(validateJobSchema(doc));
 
-        // Workflow
-        doc.workflow = generateWorkflow(doc);
-        errors = errors.concat(validateWorkflow(doc));
+        // Workflow. Remove this after we stop supporting workflow
+        if (doc.workflow || !hasRequires) {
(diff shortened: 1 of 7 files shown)


lib/phase/functional.js
```diff
@@ -181,14 +181,17 @@ function validateWorkflow(doc) {
 module.exports = flattenedDoc => (
     new Promise((resolve, reject) => {
         const doc = flattenedDoc;
+        const hasRequires = Object.keys(doc.jobs).some(jobName => doc.jobs[jobName].requires);
         let errors = [];
 
         // Jobs
         errors = errors.concat(validateJobSchema(doc));
 
         // Workflow
```
tkyi: Maybe it'd be helpful to have comments like `legacy workflow` or something for code that we'll need to remove

lib/phase/permutation.js
```diff
@@ -37,6 +37,10 @@ module.exports = validatedDoc => (
                 job.description = doc.jobs[jobName].description;
             }
 
+            if (doc.jobs[jobName].requires !== undefined) {
```
tkyi: Probably want to add a comment saying `Check for new config`

d2lam: Added comments

lib/phase/functional.js
d2lam: Added some comments

jithine:
@d2lam build failure. otherwise 👍

d2lam:
@jithin1987  Thanks! Build failed because it needs the new data-schema to be merged first :)

PR #48, opened 2017-10-10
feat: add workflowGraph to parsed config

minzcmu:
- add `workflowGraph` to parsed config for both legacy and new workflow
- move logic to set `requires` to an array to the `flatten` part since `workflow-parser` assumes `requires` is an array

Block by [workflow-parser](https://github.com/screwdriver-cd/workflow-parser), after that I need to fix some tests


lib/phase/functional.js
```diff
@@ -141,6 +142,29 @@ function generateWorkflow(doc) {
 }
 
 /**
+ * Get the defined workflow, or if missing generate one (list of all jobs in series)
+ * @method generateWorkflow
+ * @param  {Object}          doc Document that went through flattening
+ * @return {Array}               Workflow of jobs
```
tkyi: Should be an object, right?

package.json
```diff
@@ -53,6 +53,7 @@
     "js-yaml": "^3.8.3",
     "keymbinatorial": "^1.0.0",
     "screwdriver-data-schema": "^18.0.0",
+    "screwdriver-workflow-parser": "^0.0.0",
```
tkyi: Should this be version `1.0.0` eventually?

petey: yes, I'll bump major for the next release.

lib/phase/functional.js
```diff
@@ -141,6 +142,29 @@ function generateWorkflow(doc) {
...
+ * @method generateWorkflowGraph
+ * @param  {Object}          doc Document that went through flattening
+ * @return {Object}              WorkflowGraph that consists of nodes and edges
+ */
+function generateWorkflowGraph(doc) {
+    // In case doc got changed during workflow parsing
+    const cloneDoc = clone(doc);
+
+    if (doc.workflow) {
+        // Remove main job since workflow-parser assumes that
```
petey: workflow parser will do this now.

lib/phase/functional.js
```diff
@@ -168,6 +192,30 @@ function validateWorkflow(doc) {
...
+    const sortedJobList = JSON.stringify(clone(Object.keys(doc.jobs)).sort());
+    const nodesNames = clone(doc.workflowGraph.nodes).map(node => node.name);
+    const filteredNodes = nodesNames.filter(name => name !== '~commit' && name !== '~pr');
+    const sortedNodes = JSON.stringify(filteredNodes.sort());
+
+    if (sortedNodes !== sortedJobList) {
+        return ['WorkflowGraph: must contain all the jobs listed in "jobs" in nodes'];
+    }
+
+    // TODO: check if there is cycle in the graph
```
petey: `WorkflowParser.hasCycle(workflowGraph) // true/false`

minzcmu: Petey, I remove the first check to see if nodes contain all jobs. The err return here is for users like they put things wrong in `yaml`. But that check is like checking the functionality of workflow-parser.

lib/phase/functional.js
```diff
@@ -193,6 +211,9 @@ module.exports = flattenedDoc => (
             errors = errors.concat(validateWorkflow(doc));
         }
 
+        doc.workflowGraph = generateWorkflowGraph(doc);
+        // TODO: validate if  doc.workflowGraph has cycle
```
tkyi: Still need to use the new method to check for cycles.

tkyi (approved):

lib/phase/flatten.js
```diff
@@ -23,7 +23,7 @@ function merge(newJob, oldJob, fromTemplate) {
     newJob.image = oldJob.image || newJob.image;
 
     if (oldJob.requires) {
-        newJob.requires = oldJob.requires;
+        newJob.requires = [].concat(oldJob.requires);
```
petey, line 26: Nice.

lib/phase/functional.js
```diff
@@ -141,6 +142,23 @@ function generateWorkflow(doc) {
...
+ */
+function generateWorkflowGraph(doc) {
+    // In case doc got changed during workflow parsing
+    const cloneDoc = clone(doc);
+
+    if (doc.workflow) {
+        return WorkflowParser.getWorkflow(cloneDoc, { useLegacy: true });
+    }
+
+    return WorkflowParser.getWorkflow(cloneDoc);
```
petey, line 158: This is a difficult call to make here. We have projects that do not define workflow but do have more than one job. The default behavior in legacy mode is to convert the list of jobs into a linear workflow. In non-legacy mode, this will be N detached pipelines.

I think what we want to do here is actually _always_ run workflow through legacy-mode, the parser will do the right thing if it detects non-legacy `requires` configs. When we are satisfied that we no longer wish to support legacy mode, we just stop passing in the `useLegacy` flag.

petey (changes requested):

lib/phase/functional.js
minzcmu: Based on the logic we have already, in that case, it will always generate the legacy workflow.
https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/functional.js#L191

So after that when it comes to `generateWorkflowGraph`, we will have the workflow already.

petey: So our legacy pipelines that do not explicitly define a workflow already have something defining a workflow before it gets here? We should consider removing those bits, since workflow parser should be handling that, but I guess we can do that later.

petey (approved):

PR #50, opened 2017-10-23
fix(723): requires main job for only legacy config

minzcmu:
We will remove `main` as required from data-schema.
Add an extra check here to validate that for legacy config. 

Blocked by https://github.com/screwdriver-cd/data-schema/pull/188


tkyi:
Probably should add a test

d2lam (approved):

tkyi (approved):

PR #51, opened 2017-10-23
fix: increase ENV limit

d2lam:
Increasing the Environment Variable limit since some jobs might need more than 25 environment variables. 

Based on @Renaud's request


cybik (approved):
Thank you!

tkyi (approved):
Seems reasonable

minzcmu (approved):
👍

PR #53, opened 2017-11-03
fix: remove error for join since we support it now

d2lam:
Remove this error since we will start supporting join trigger(https://github.com/screwdriver-cd/screwdriver/issues/770)


minzcmu (approved):
👍

PR #59, opened 2018-04-05
fix: add env variables to indicate template name and version

minzcmu:
## Context
Users may want to know the exact template version inside their builds for debug purpose

## Objective
This PR will make `SD_TEMPLATE_NAME` and `SD_TEMPLATE_VERSION` available inside users' build if they are using templates.

Related to: https://github.com/screwdriver-cd/screwdriver/issues/960

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -133,6 +133,13 @@ function mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory) {
             }
 
             const newJob = template.config;
+            const environment = newJob.environment || {};
+
+            // Inject template version and name to env
+            newJob.environment = Hoek.merge(environment, {
+                SD_TEMPLATE_VERSION: template.version,
+                SD_TEMPLATE_NAME: template.name
+            });
 
             merge(newJob, oldJob, true);
             delete newJob.template;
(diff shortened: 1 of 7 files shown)


lib/phase/flatten.js
```diff
@@ -133,6 +133,13 @@ function mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory) {
             }
 
             const newJob = template.config;
+            const environment = newJob.environment || {};
+
+            // Inject template version and name to env
+            newJob.environment = Hoek.merge(environment, {
```
d2lam: `Hoek.merge` is destructive, it modifies the first parameter itself.

d2lam (approved):
LGTM 👍

lib/phase/functional.js
```diff
@@ -28,7 +28,9 @@ function validateJobMatrix(job, prefix) {
     const environment = Hoek.reach(job, 'environment', {
         default: {}
     });
-    const environmentSize = Object.keys(matrix).length + Object.keys(environment).length;
+    const SDEnv = ['SD_TEMPLATE_NAME', 'SD_TEMPLATE_VERSION'];
```
tkyi: Is there any way to convert this section into a method? Since it seems to duplicate logic done at line 116
Or at least declare these vars at a more global level

tkyi (approved):
Small comment, otherwise LGTM 👍

lib/phase/functional.js
minzcmu: Makes sense! will make the change

PR #61, opened 2018-04-18
fix(hip): Upgrade to node:8

philipjscott:
## Context

Node.js 6 is slowly becoming deprecated, with popular packages such as `semantic-release` no longer supporting it.

## Objective

* Change image in `screwdriver.yaml` to `node:8`
* Update the `engines` property in `package.json`

### DIFF
--- a/test/data/node-module.yaml
+++ b/test/data/node-module.yaml
@@ -1,6 +1,3 @@
-workflow:
-    - publish
-
 shared:
     environment:
         NODE_ENV: test
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     }
   },
   "engines": {
-    "node": "^6.0.0"
+    "node": "^8.0.0"
   },
   "engine-strict": true,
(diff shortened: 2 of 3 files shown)


Filbird (approved):

test/data/node-module.yaml
```diff
@@ -1,6 +1,3 @@
-workflow:
```
minzcmu, line 1: You need requires: [~pr, ~commit] for the main job and [main] for publish job

minzcmu (changes requested):

test/data/node-module.yaml
philipjscott: Isn't the `requires` already there?

https://github.com/screwdriver-cd/config-parser/blob/node8/test/data/node-module.yaml#L13
https://github.com/screwdriver-cd/config-parser/blob/node8/test/data/node-module.yaml#L27

minzcmu (approved):

PR #62, opened 2018-05-08
feat(1041): Support job scm

Filbird:
Include the `scm` object as part of the parsed config. Similar to other fields, job-level `scm` will overwrite shared-level `scm`.

https://github.com/screwdriver-cd/screwdriver/issues/1041


Filbird:
Closing since `SCM` will not be a job-level configuration

PR #67, opened 2018-06-05
feat(1040): resolve job image alias with template images

jimgrund:
## Context

If a user job references an image alias from a template images directive, this should be flattened such that the job uses the resolved image path.

## References

* Issue: https://github.com/screwdriver-cd/screwdriver/issues/1040

### DIFF
--- a/test/data/basic-job-with-images.yaml
+++ b/test/data/basic-job-with-images.yaml
@@ -1,7 +1,7 @@
 jobs:
     main:
         image: latest-image
-        template: imagestemplate@2
+        template: ImagesTestNamespace/imagestemplate@2
         description: "This is the main description"
     publish:
         image: node:4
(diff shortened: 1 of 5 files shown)


test/data/templateWithImages.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "imagestemplate",
```
tkyi: Since all templates should have a namespace now, can you add one to this example either by adding a `namespace` field or using `name: namespace/imagestemplate`?

jithine (commented):
LGTM once Tiff's review is addressed.

test/data/basic-job-with-images.yaml
```diff
@@ -0,0 +1,10 @@
+jobs:
+    main:
+        image: latest-image
+        template: imagestemplate@2
```
tkyi: should be `template: ImagesTestNamespace/imagestemplate@2`

tkyi (approved):

PR #69, opened 2018-06-26
feat: allow user-teardown-steps

d2lam:
Enforce `user-teardown-steps` to be at the end
Related: https://github.com/screwdriver-cd/screwdriver/issues/1126

### DIFF
--- a/lib/phase/structural.js
+++ b/lib/phase/structural.js
@@ -4,6 +4,37 @@ const Joi = require('joi');
 const SCHEMA_CONFIG = require('screwdriver-data-schema').config.base.config;
 
 /**
+ * Check if user teardown is at the end of the yaml
+ * @method checkUserTeardown
+ * @param  {Object}          data Screwdriver yaml
+ * @return {Promise}              Error if there is an error
+ */
+function checkUserTeardown(data) {
+    let steps = [];
+    let firstUserTeardownStep = -1;
+    let error = null;
+
+    Object.keys(data.jobs).forEach((job) => {
+        steps = data.jobs[job].steps;
+
+        if (steps) { // if there are user-defined steps
+            for (let i = 0; i < steps.length; i += 1) {
+                const isUserTeardown = Object.keys(steps[i])[0].startsWith('user-teardown-');
+
+                if (firstUserTeardownStep > -1 && !isUserTeardown) {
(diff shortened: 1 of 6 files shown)


lib/phase/structural.js
```diff
@@ -3,6 +3,37 @@
...
+    let steps = [];
+    let firstUserTeardownStep = -1;
+    let error = null;
+
+    Object.keys(data.jobs).forEach((job) => {
+        steps = data.jobs[job].steps;
+
+        if (steps) { // if there are user-defined steps
+            for (let i = 0; i < steps.length; i += 1) {
+                const isUserTeardown = Object.keys(steps[i])[0].startsWith('user-teardown-');
```
jithine: can we just call it `teadown-` ?  Why make it more verbose with `user-` prefix.

d2lam: Sure I can change that

lib/phase/structural.js
```diff
@@ -3,6 +3,37 @@
...
+        if (steps) { // if there are user-defined steps
+            for (let i = 0; i < steps.length; i += 1) {
+                const isUserTeardown = Object.keys(steps[i])[0].startsWith('teardown-');
+
+                if (firstUserTeardownStep > -1 && !isUserTeardown) {
+                    error = new Error('User teardown steps need to be at the end');
+                }
+
+                if (isUserTeardown) {
+                    firstUserTeardownStep = i;
```
minzcmu: Based on the logic, this `firstUserTeardownStep` seems to be `lastUserTeardownStep`? It will get updated to `teardown-2` if you have 
```yaml
- teardown-1
- teardown-2
```

d2lam: It doesn't matter if you have
```
teardown1
teardown2
```
The intention here is if there is a already a `teardown` and there is another `non-teardown` afterward then it will be error. Otherwise just keep proceeding. I can change the variable name to something else.

tkyi (approved):

lib/phase/structural.js
minzcmu: I feel it's better we change the name to something else, probably without `first`. Otherwise it's kinda confusing

minzcmu (approved):

PR #76, opened 2018-10-16
feat(1280): support cache config

d2lam:
Flatten cache settings to each job. 
If there is a setting for a non-existing job, will return error. 

Blocked by:
https://github.com/screwdriver-cd/data-schema/pull/287

Related: 
https://github.com/screwdriver-cd/screwdriver/issues/1257
https://github.com/screwdriver-cd/screwdriver/issues/1280


pranavrc (approved):
image

PR #79, opened 2018-11-16
feat(1319): Throw error if build cluster annotation is not a valid option

tkyi:
## Context
In order to support multiple build clusters, we need to update the validator.

## Objective
This PR adds logic to check if the build cluster specified in the `screwdriver.cd/buildCluster` annotation is actually a valid buildCluster in the database.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1319

### DIFF
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -20,6 +20,36 @@ const SDEnv = [
 ];
 
 /**
+ * Ensure the build cluster annotation is valid
+ * @method validateBuildClusterAnnotation
+ * @param  {BuildClusterFactory}  buildClusterFactory Build cluster Factory to get build clusters
+ * @param  {Object}               doc Document that went through flattening
+ * @return {Array}                List of errors
+ */
+function validateBuildClusterAnnotation(doc, buildClusterFactory) {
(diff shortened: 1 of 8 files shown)


lib/phase/functional.js
```diff
@@ -19,6 +19,36 @@ const SDEnv = [
...
+/**
+ * Ensure the build cluster annotation is valid
+ * @method validateBuildClusterAnnotation
+ * @param  {BuildClusterFactory}  buildClusterFactory Build cluster Factory to get build clusters
+ * @param  {Object}               doc Document that went through flattening
+ * @return {Array}                List of errors
+ */
+function validateBuildClusterAnnotation(doc, buildClusterFactory) {
+    const buildClusterAnnotation = 'screwdriver.cd/buildCluster';
+    const buildClusterName = Hoek.reach(doc, 'annotations') ?
```
d2lam: You can do something like this too:
```
buildClusterName = hoek.reach(doc, `annotations>${buildClusterAnnotation}`, { separator: '>'})
```

d2lam (approved):

PR #89, opened 2019-09-12
fix: Should evaluate environment variables by declaration order

tkyi:
## Context

Sometimes users override environment variables previously defined in templates. While the environment variable would be overwritten, the order in which the environment variables was evaluated stayed the same. This can cause issues if the environment variable they override depends on another variable defined in their screwdriver.yaml file.

**For example:** 

Template has `[ A=, B=2 ]`
Shared has `[X=1, A=X]`
Consequently, the order of environment variables => `[A=X, B=2, X=1]`

## Objective

This PR checks if the environment variable being declared already exists. If so, it will delete it. Then it will do the merge.

## License


lib/phase/flatten.js
```diff
@@ -68,10 +68,20 @@ function merge(newJob, oldJob, fromTemplate) {
...
 
+    // Handle duplicate environment variables
+    if (newJob.environment && oldJob.environment) {
+        Object.keys(newJob.environment).forEach((newEnv) => {
+            if (newEnv in oldJob.environment) {
+                delete newJob.environment[newEnv];
+            }
+        });
+        // Merge environment variables
+        Object.assign(newJob.environment, oldJob.environment || {});
```
DekusDenial, line 82: this line alone should get the job done right? `Object.assign(A, B)` whatever in B override whatever in A, shallowly. Is there other reason why you wanna mutate `newJob` here during the merge? I dont expect this merge to be destructive right

tkyi: I think I explained in the example above

Sometimes the template might define a variable `A`, which is put into the array first `[A]`
but if the user defines a variable `B` in their job and also defines `A=B`, the environment variables will not be evaluated properly since the array will be `[A=B, B]` based on the current implementation. `B` will not exist before `A` is evaluated, and `A` would appear to be undefined.

This PR would delete any overlapping vars so it can be evaluated later => `[B, A=B]`

test/data/basic-job-with-environment.json
```diff
@@ -0,0 +1,80 @@
...
+        "environment": {
+          "A": "meow",
+          "B": "${A}",
+          "BAR": "foo",
+          "BAZ": "world",
+          "FOO": "${B}",
+          "SD_TEMPLATE_FULLNAME": "mynamespace/environment",
+          "SD_TEMPLATE_NAME": "environment",
+          "SD_TEMPLATE_NAMESPACE": "mynamespace",
+          "SD_TEMPLATE_VERSION": "1.2.3"
```
jithine: this is alphabetical, different from expected...

lib/phase/flatten.js
jithine: simply overriding A is not enough, order should come from B. These are shell environment variables. But I am puzzled by test job json output. GET /jobs/:jobID seems to imply that it's insertion order for environment variables object (JS guarantees in various implementations, though I am unsure of actual spec), but test result suggests it's alphabetical.

DekusDenial: now that i understand the problem, but based on my comment on slack, i think the problem is bigger than this and can be easily surfaced w/ this example:
```yaml
shared:
    A: 1
    B: 2
    C: $A
job:
    A: 4
```
after going thru the code above which simply translates to ` (shared - (shared ∩ job)) ∪ job`, and based on the insertion order of `job`, which is just append, u will end up with:
```yaml
B: 2
C: $A
A: 4
```
and very likely the problem will be more obvious if `C` depends on both A and B.
Therefore, this is really a toposort problem at heart

jithine: good catch, merely maintaining order is not enough to cover all cases.
Now this is hard to pull of, because values can be just strings. lets say `C: "$a/something"`

jithine: what if we end up something like

```
A: 1
B: 2
C: $A
A: 4
```

shell ultimately processes variable declarations in order. We could leave it to the user to override further dependencies, well it will be awkward and verbose, ~but will simplify implementation~, and will lead to data schema changes. 

So making toposort work will be our best bet.

DekusDenial: you can document it somewhere and let people to understand the correct usage. e.g. whether to accept env var expansion at the shared context, or people will have to copy the env var which has expansion with some of their custom defined ones in the job context, and  other options, just needa communicate precisely of what is expected

jithine (approved):

ameir:
I don't know how the environment is ultimately passed to the jobs, but this can be solved if the environment is internally treated as an array.  An `env` array can be initialized with the elements of `newJob.environment`, and oldJob.environment can be appended to it.  This will keep ordering, including re-used keys.  When the variables are exported, they will be evaluated in order.

Simple example since I don't know how the variables are injected into the job:
Template has [ A=1, B=2, C="$A/something" ]
Shared has [X=1, A=X]

Internal array is: ```[{"A":1}, {"B":2}, {"C":"$A/something"}, {"X":1}, {"A": "$X"}]```

Shell evaluates like so:
```
export A=1
export B=2
export C=$A/something
export X=1
export A=$X
```
Note that `C` will depend on the original `A`, but I think that's just a documentation point.

DekusDenial:
yes, the intention would be ambiguous. Having a better doc will save people time and avoid situation like this. Once we explain how we process env vars among contexts, it's up to the user to interpret following the rules.

ameir:
Based on https://github.com/screwdriver-cd/launcher/blob/master/launch.go#L470, this solution would work.  A toposort would be great, albeit a much greater level of effort, and you'd probably have to enforce a certain syntax like `${var}`, else you'll have to understand all the ways that a shell can parse variables.

PR #95, opened 2019-10-29
fix(1722): enable to use >= node8

tk3fftk:
## Context

We should update to Node.js version. But this package may prevent updating.

## Objective

This PR enables to use >= node 8.16.0 which is latest version of node 8.x.

## References
- https://github.com/screwdriver-cd/screwdriver/issues/1722
- https://nodejs.org/dist/latest-v8.x/

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,14 +37,14 @@
     }
   },
   "engines": {
-    "node": "^8.0.0"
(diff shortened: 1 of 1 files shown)


adong (approved):
LGTM

PR #101, opened 2020-04-23
feat(1748): [2] Check job cache

kumada626:
## Context

If job's cache property is set false, that job should not have cache config.

## Objective

Does not set cache if job's cache property is false.

Test fails until screwdriver-cd/data-schema#394 is merged.

## References

screwdriver-cd/screwdriver#1748
screwdriver-cd/data-schema#394

## License

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -46,8 +46,12 @@ function flattenCacheSettings(cacheConfig, jobs) {
     };
 
     Object.keys(jobs).forEach((jobName) => {
-        jobs[jobName].cache = Hoek.applyToDefaults(cache,
-            { job: Hoek.reach(cacheConfig, `job.${jobName}`, { default: [] }) });
+        if (jobs[jobName].cache !== false) {
+            jobs[jobName].cache = Hoek.applyToDefaults(cache,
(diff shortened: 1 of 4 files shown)


lib/phase/flatten.js
```diff
@@ -46,8 +46,12 @@ function flattenCacheSettings(cacheConfig, jobs) {
     };
 
     Object.keys(jobs).forEach((jobName) => {
-        jobs[jobName].cache = Hoek.applyToDefaults(cache,
-            { job: Hoek.reach(cacheConfig, `job.${jobName}`, { default: [] }) });
+        if (jobs[jobName].cache !== false) {
```
wahapo, line 49: I'm not sure, if the `cache` isn't specified, shouldn't the error occur?

yoshwata (approved):

lib/phase/flatten.js
kumada626: Not error if cache is not specified. If cache is specified, cache is evaluated as undefined. undefined !== false returns true.

wahapo: ok:+1:

wahapo (approved):

ibu1224 (approved):

tkyi (approved):

PR #102, opened 2020-05-08
fix: use ANSI-C quoting for echoing error

DekusDenial:
## Context

double quote is bad when you can inject command

## Objective

use ANSI-C quoting for escaping 

## License


tkyi (approved):

s4r4r:
@tolar fyi

PR #108, opened 2020-08-04
feat(1958): Hapi v19 dependencies upgrade. BREAKING CHANGE: v19.x.x

pritamstyz4ever:
## Context

Screwdriver hapi.js dependencies are outdated and not supported anymore.

## Objective

This PR upgrades hapi and node fixes all schema related changes to be compatible with latest @hapi js dependencies.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1958

## License

### DIFF
--- a/lib/phase/structural.js
+++ b/lib/phase/structural.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const Joi = require('joi');
 const SCHEMA_CONFIG = require('screwdriver-data-schema').config.base.config;
 
 /**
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -48,13 +48,13 @@
...
     "clone": "^2.1.2",
-    "hoek": "^5.0.4",
-    "joi": "^13.7.0",
+    "joi": "^17.1.1",
     "js-yaml": "^3.13.0",
     "keymbinatorial": "^1.1.5",
-    "screwdriver-data-schema": "^19.1.1",
-    "screwdriver-workflow-parser": "^2.0.0",
+    "screwdriver-data-schema": "git://github.com/screwdriver-cd/data-schema.git#hapi-v19",
+    "screwdriver-workflow-parser": "git://github.com/screwdriver-cd/workflow-parser.git#hapi-v19",
```
tkyi: reminder to update before merge

test/index.test.js
```diff
@@ -124,7 +124,7 @@ describe('config parser', () => {
                 () => parser(loadData('bad-step-name.yaml'))
                     .then((data) => {
                         assert.match(data.jobs.main[0].commands[0].command,
-                            /"foo bar" only supports the following characters A-Z,a-z,0-9,-,_/);
+                            /foo bar" only supports the following characters A-Z,a-z,0-9,-,_/);
```
tkyi: is this weird? why is it missing a leading quotation mark?

pritamstyz4ever: The complete message is actually [Validation Error] "jobs.main.steps[0].foo bar" ..., so we are matching the pattern.

lib/phase/structural.js
```diff
@@ -49,22 +48,22 @@ function checkAdditionalRules(data) {
...
+        if (errors) {
+            throw errors;
+        }
 
-            return resolve(data);
-        });
-    })
-);
+        return data;
+    } catch (err) {
```
jithine: why is this catch needed ? can't the entire `try..catch` be omitted and let the function throw err

test/index.test.js
jithine: shouldn't it have matched even with the `"`?

pritamstyz4ever: Actual error message now includes the entire object structure.
ValidationError: "jobs.main.steps[0].foo bar" only supports the following characters A-Z,a-z,0-9,-,_
which is not a correct regex in itself. So made a limited change to message.

jithine: missed that

jithine (approved):

tkyi (approved):

PR #112, opened 2020-10-22
fix(2211): Error on validating sd.yaml that has cache-disable setting

### LINKED ISSUE #2211: Incorrect cache config in job  breaks pipeline dispay
**What happened**:

A pipeline with following config will result in 404 in UI. This is because `/jobs` end point returns 500 with validation error

```
jobs:
  staging-trigger:
    requires: [~commit:develop]
    cache: false
    image: ruby:2.7.1
    sourcePaths: ["screwdriver.yaml", "Dockerfile"]
    steps:
      - noop: echo "noop"
```

**What you expected to happen**:

GET to `/jobs` should not have failed. Screwdriver should not have created the incorrect job in first place or made it in a way that not having global cache config does not break `/jobs/` end point

**How to reproduce it**:

Create a pipeline with above yaml config.

yoshwata:
## Context

fix https://github.com/screwdriver-cd/screwdriver/issues/2211

When the object base cache settings does not exist, the cache-disable settings (true/false) keep remains as shape of boolean. It is validated as [output](https://github.com/screwdriver-cd/screwdriver/blob/e698e39916dc84922f3c0e21563cf10c4a379669/plugins/validator.js#L40) of parser, but it is not match with the shape of [data-schema](https://github.com/screwdriver-cd/data-schema/blob/3c7ba87df825a47291454c955782c61644599158/api/validator.js#L27).

This PR fixes the validation error by deleting data which is no need to exists for validation. 

## Objective

The 500 error does not occur even if cache-disable settings(true/false) exists without cache settings(object)

## References

https://github.com/screwdriver-cd/screwdriver/issues/2211

## License

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -374,6 +374,10 @@ module.exports = (parsedDoc, templateFactory) => {
     if (parsedDoc.cache) {
         doc.jobs = flattenCacheSettings(parsedDoc.cache, doc.jobs);
         delete doc.cache;
+    } else {
+        Object.keys(doc.jobs).forEach((jobName) => {
+            delete doc.jobs[jobName].cache;
+        });
     }
 
     // Flatten templates
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -376,7 +376,9 @@ module.exports = (parsedDoc, templateFactory) => {
         delete doc.cache;
     } else {
         Object.keys(doc.jobs).forEach((jobName) => {
-            delete doc.jobs[jobName].cache;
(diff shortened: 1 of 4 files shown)


tkyi (approved):

wahapo (approved):

kumada626 (approved):

jithine (approved):
Is there any way to surface up this error to use ?

yoshwata:
@jithine I fixed to show like below.
Screen Shot 2020-10-23 at 16 56 05

jithine (approved):

PR #113, opened 2020-10-29
feat(2174): [2] Merge steps with annotations

kumada626:
## Context

If annotation `screwdriver.cd/mergeStep` is true, steps defined in shared field are not overridden but merged with job defined steps.
This feature only enable when template is used.

The test will pass after [data-schema](https://github.com/screwdriver-cd/data-schema/pull/418) is merged.

## References

issue: screwdriver-cd/screwdriver#2174
data-schema: https://github.com/screwdriver-cd/data-schema/pull/418

## License

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -230,8 +230,8 @@ function mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory, shar
 
             // merge shared steps into oldJob
             if (sharedConfig.steps
-                && (oldJob.annotations['screwdriver.cd/mergeStep']
-                || template.annotations['screwdriver.cd/mergeStep'])) {
+                && (oldJob.annotations['screwdriver.cd/mergeSharedSteps']
+                || template.annotations['screwdriver.cd/mergeSharedSteps'])) {
                 // convert steps from oldJob from array to object for faster lookup
                 const oldSteps = oldJob.steps.reduce((obj, item) => {
                     const key = Object.keys(item)[0];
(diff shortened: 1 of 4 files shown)


lib/phase/flatten.js
```diff
@@ -228,6 +228,26 @@ function mergeTemplateIntoJob(jobName, jobConfig, newJobs, templateFactory) {
                 SD_TEMPLATE_VERSION: template.version
             });
 
+            // merge shared steps into oldJob
+            if (sharedConfig.steps
+                && (oldJob.annotations['screwdriver.cd/mergeStep']
```
wahapo: I think `mergeSharedSteps` is easier for users to understand. It may not be necessary considering that the only steps to merge are the values that defined in `shared`, but the user needs to fully understand how it works to use them, whereas with `mergeSharedSteps`, the user doesn't have to think about it at all.

test/data/basic-job-with-shared-and-template-override-steps.json
```diff
@@ -0,0 +1,57 @@
+{
+    "annotations": {},
+    "parameters": {},
+    "jobs": {
+        "main": [{
+            "annotations": {
+                "screwdriver.cd/deepMergeStep": true
```
wahapo: typo?

lib/phase/flatten.js
kumada626: Change annotations to `mergeSharedSteps`.

test/data/basic-job-with-shared-and-template-override-steps.json
kumada626: Fixed.

wahapo (commented):
LGTM If test passes.

tkyi (approved):

PR #119, opened 2021-02-03
feat(2319): Support order for template composition

tkyi:
## Context

Would be nice if users could pick and choose from steps defined in their own screwdriver.yamls and template configs to determine which commands run.

## Objective

This PR adds support for `order` keyword and warnings.

Notes:
- `order` keyword can only be used when `template` is used.
- job-level step definitions will take priority over template step definitions
- job-level step definitions will take priority over shared-level step definitions. These combined will take priority over template step definitions when [screwdriver.cd/mergeSharedSteps: true](https://docs.screwdriver.cd/user-guide/templates#merging-with-shared-steps)
- implicit [wrap (pre/post)](https://docs.screwdriver.cd/user-guide/templates#wrap) function is not supported with `order` keyword.
- implicit [teardown-](https://docs.screwdriver.cd/user-guide/configuration/jobconfiguration#teardown) steps will still work

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2319

## License

### DIFF
--- a/index.js
+++ b/index.js
@@ -94,6 +94,32 @@ function validateReservedAnnotation(doc) {
 }
 
 /**
+ * Check that order is only used with template
+ * @method validateOrderAndTemplate
+ * @param  {Object}           doc               Document that went through structural parsing
+ * @return {Array}                              List of warnings
+ */
+function validateOrderAndTemplate(doc) {
+    let warnings = [];
+
+    let template = Hoek.reach(doc.shared, 'template');
+
+    Object.keys(doc.jobs).forEach((jobName) => {
+        const order = Hoek.reach(doc.jobs[jobName], 'order', { default: [] });
+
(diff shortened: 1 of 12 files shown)


index.js
```diff
@@ -154,8 +180,14 @@ module.exports = function configParser(
         // Flatten structures
         .then((parsedDoc) => {
             warnMessages = warnMessages.concat(validateTemplateVersion(parsedDoc, templateFactory));
+            warnMessages = warnMessages.concat(validateOrderAndTemplate(parsedDoc));
```
jithine: is this separate fn needed ? can't `phaseFlatten` handle it ? it's anyways doing validation for step existence ?

tkyi: updated

jithine (approved):

PR #125, opened 2021-06-10
fix: try to speed up coverage step by increasing memory

tkyi:
## Context

Noticed that the coverage step takes >10mins on average for the config-parser pipeline.

## Objective

This PR increases memory to see if that helps speed up that step.

## References

Screen Shot 2021-06-10 at 3 42 00 PM

## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -5,6 +5,7 @@ jobs:
     main:
         environment:
             SD_SONAR_OPTS: "-Dsonar.sources=index.js,lib -Dsonar.tests=test -Dsonar.javascript.lcov.reportPaths=artifacts/coverage/lcov.info -Dsonar.testExecutionReportPaths=artifacts/report/test.xml"
+            NODE_OPTIONS: "--max_old_space_size=4096"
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -5,6 +5,7 @@ jobs:
     main:
         environment:
             SD_SONAR_OPTS: "-Dsonar.sources=index.js,lib -Dsonar.tests=test -Dsonar.javascript.lcov.reportPaths=artifacts/coverage/lcov.info -Dsonar.testExecutionReportPaths=artifacts/report/test.xml"
+            NODE_OPTIONS: "--max_old_space_size=4096"
```
jithine: seems to be a sharp jump from 500MB default. Also sonar-scanner is a java based binary, don't think setting is going to affect it's performance

jithine: from what I recall, more CPU to build seems to improve it's performance 🤔

tkyi: updated

jithine (approved):

PR #143, opened 2022-12-21
fix(2767): Add support for stage setup and teardown [3]

tkyi:
## Context

Stages might want to always run initialization or cleanups needed before jobs in a stage can be executed.

## Objective

This PR converts stage setup/teardown into jobs in parser.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2767

## License


lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+                    // Handle startFrom
+                    if (newJobs[job].requires && newJobs[job].requires.includes(`~stage@${stageName}:setup`)) {
+                        newJobs[job].stage.startFrom = true;
+                        stages[stageName].startFrom = job;
+                        delete newJobs[job].requires;
+                    }
+                }
+            });
+
+            if (!Hoek.deepEqual(setup, {})) {
```
sagar1312: Shouldn't setup job be implicitly added if it is not specified in the configuration?

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+                newJobs[`${STAGE_PREFIX}${stageName}:setup`] = setup;
+                newJobs[`${STAGE_PREFIX}${stageName}:setup`].stage = { name: stageName };
+            }
+            delete stages[stageName].setup;
+            if (!Hoek.deepEqual(teardown, {})) {
```
sagar1312: Shouldn't `teardown` job be implicitly added if it is not specified in the configuration?

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+        Object.keys(stages).forEach(stageName => {
+            const { setup, teardown, jobs: stageJobs } = stages[stageName];
+
+            stageJobs.forEach(job => {
+                if (newJobs[job]) {
+                    newJobs[job].stage = { name: stageName };
+                    // Handle startFrom
+                    if (newJobs[job].requires && newJobs[job].requires.includes(`~stage@${stageName}:setup`)) {
+                        newJobs[job].stage.startFrom = true;
+                        stages[stageName].startFrom = job;
```
sagar1312: Per this comment https://github.com/screwdriver-cd/data-schema/pull/510#discussion_r1167107851, we could have more than one job with in a stage that needs to get triggered after `setup` is completed.

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+                        delete newJobs[job].requires;
```
sagar1312: Why do we need to delete `requires` from `job` configuration?
To support a scenario like https://github.com/screwdriver-cd/data-schema/pull/510#discussion_r1186392713, `requires` may include other jobs besides `setup` and deleting `requires` could lead to unexpected workflow.

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+    if (stages && !Hoek.deepEqual(stages, {})) {
+        Object.keys(stages).forEach(stageName => {
+            const { setup, teardown, jobs: stageJobs } = stages[stageName];
+
+            stageJobs.forEach(job => {
+                if (newJobs[job]) {
+                    newJobs[job].stage = { name: stageName };
+                    // Handle startFrom
+                    if (newJobs[job].requires && newJobs[job].requires.includes(`~stage@${stageName}:setup`)) {
+                        newJobs[job].stage.startFrom = true;
```
sagar1312: We do not have `stage` in `job` domain and/or DB model. What is the need for this derived attribute?

Can we rely on the presence of `setup` job name in `requires` wherever (in the workflow parser) needed to determine if the needs to be triggered after the execution of `setup` job.

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
 
+/**
+ * Flatten stage setup and teardown into jobs object
+ * @param  {Object} doc               Document that went through structural parsing
+ */
+function flattenStageSetupAndTeardownJobs(doc) {
+    const newJobs = clone(doc.jobs);
+    const stages = clone(doc.stages);
+
+    if (stages && !Hoek.deepEqual(stages, {})) {
```
sagar1312: When would `stages` be an empty object `{}`?
Shouldn't it be `null`/`undefined` if `stages` is not defined in the `screwdriver.yaml`?

lib/phase/flatten.js
sagar1312: When would `setup` be an empty object `{}`?

lib/phase/flatten.js
sagar1312: (Similar to `setup`) When would `teardown` be an empty object `{}`?

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+            delete stages[stageName].setup;
+            if (!Hoek.deepEqual(teardown, {})) {
+                newJobs[`${STAGE_PREFIX}${stageName}:teardown`] = teardown;
+                newJobs[`${STAGE_PREFIX}${stageName}:teardown`].stage = { name: stageName };
+            }
+            delete stages[stageName].teardown;
+        });
+    }
+
+    return { jobsWithStages: newJobs, newStages: stages };
```
sagar1312:
```suggestion
    return { jobs: newJobs, stages: stages };
```

lib/phase/flatten.js
```diff
@@ -522,6 +537,46 @@ function convertSteps(jobs) {
...
+ */
+function flattenStageSetupAndTeardownJobs(doc) {
+    const newJobs = clone(doc.jobs);
+    const stages = clone(doc.stages);
+
+    if (stages && !Hoek.deepEqual(stages, {})) {
+        Object.keys(stages).forEach(stageName => {
+            const { setup, teardown, jobs: stageJobs } = stages[stageName];
+
+            stageJobs.forEach(job => {
```
sagar1312:
```suggestion
            stageJobs.forEach(jobName => {
```

lib/phase/permutation.js
```diff
@@ -64,6 +64,10 @@ module.exports = async validatedDoc => {
             job.freezeWindows = doc.jobs[jobName].freezeWindows;
         }
 
+        if (doc.jobs[jobName].stage !== undefined) {
```
sagar1312, line 67: Why do we need reference to a `stage` from a `job` and persist that in DB?

lib/phase/functional.js
```diff
@@ -254,12 +254,29 @@ module.exports = async ({
...
     let errors = [];
     let warnings = [];
+    let pipelineOnly = true;
 
     // Jobs
     errors = errors.concat(validateJobSchema(doc));
 
+    if (doc.stages) {
+        pipelineOnly = false;
+    }
```
sagar1312: Can we abstract this check into workflow parser (i.e. `generateWorkflowGraph`) as the entire doc is being passed into it?

lib/phase/functional.js
```diff
@@ -254,12 +254,29 @@ module.exports = async ({
...
+
     // Workflow graph
-    doc.workflowGraph = await generateWorkflowGraph(doc, triggerFactory, pipelineId);
+    const workflowGraphs = await generateWorkflowGraph(doc, triggerFactory, pipelineId, pipelineOnly);
```
sagar1312: We should always keep the type and structure of the return value for a function irrespective of the arguments being passed.
```suggestion
    const {pipelineWorkflow, stageWorkflows} = await generateWorkflowGraph(doc, triggerFactory, pipelineId, pipelineOnly);
```

lib/phase/functional.js
```diff
@@ -254,12 +254,29 @@ module.exports = async ({
...
+
+    if (pipelineOnly) {
+        doc.workflowGraph = workflowGraphs;
+    } else {
+        doc.workflowGraph = workflowGraphs.workflow;
+
+        if (workflowGraphs.stageWorkflows) {
+            Object.keys(workflowGraphs.stageWorkflows).forEach(stage => {
```
sagar1312:
```suggestion
            Object.keys(workflowGraphs.stageWorkflows).forEach(stageName => {
```

sagar1312 (changes requested):

lib/phase/flatten.js
```diff
@@ -105,6 +107,10 @@ function merge(newJob, oldJob, fromTemplate) {
         newJob.provider = oldJob.provider || {};
     }
 
+    if (oldJob.stage) {
```
sagar1312, line 111: Related to https://github.com/screwdriver-cd/config-parser/pull/143/files#r1242377069

lib/phase/permutation.js
tkyi: this wouldn't be persisted in the DB
just passed from config-parser to models temporarily

sagar1312: It is getting persisted in DB for all the stage jobs.

Sample pipeline
https://github.com/sagar1312/screwdriver-config/blob/main/stage-with-parallel-jobs/screwdriver.yaml

Here are the entries from `jobs` table for this pipeline config
```
+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|name             |permutations                                                                                                                                                                                                                                                      |
+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|main             |[{"image":"node:12","commands":[{"name":"config-parse-error","command":"echo 'ValidationError: \"jobs.prod-green-deploy.prod-green-test\" is not allowed. \"jobs.prod-green-deploy.prod-green-certify\" is not allowed'; exit 1"}],"secrets":[],"environment":{}}]|
|component        |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["~pr","~commit"]}]                                                                                              |
|ci-deploy        |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["publish"],"stage":{"name":"integration"}}]                                                                     |
|publish          |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["component"]}]                                                                                                  |
|ci-certify       |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["ci-test"],"stage":{"name":"integration"}}]                                                                     |
|ci-test          |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["ci-deploy"],"stage":{"name":"integration"}}]                                                                   |
|prod-blue-deploy |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["ci-certify"],"stage":{"name":"production-blue"}}]                                                              |
|prod-blue-test   |[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["prod-blue-deploy"],"stage":{"name":"production-blue"}}]                                                        |
|prod-blue-certify|[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["prod-blue-test"],"stage":{"name":"production-blue"}}]                                                          |
|prod-green-deploy|[{"annotations":{},"commands":[{"name":"init","command":"echo 'init'"}],"environment":{},"image":"node:14","secrets":[],"settings":{},"requires":["ci-certify"],"stage":{"name":"production-green"}}]                                                             |
+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

```

lib/phase/flatten.js
tkyi: sure, removed

lib/phase/flatten.js
tkyi: fixed

lib/phase/flatten.js
tkyi: it comes from data-schema since it's expecting setup/teardown object

lib/phase/flatten.js
tkyi: it comes from data-schema since it's expecting setup/teardown object

lib/phase/flatten.js
tkyi: I think it's nice/quicker way to know if a job belongs to a stage instead of looking it up every time

lib/phase/permutation.js
tkyi: ah I meant as its own field
but yeah looks like it persists in permutations blob

I can probably remove the startFrom object but the stage name tag is nice to have for quick lookup

lib/phase/flatten.js
tkyi: removed

lib/phase/flatten.js
tkyi: needs to be added here

lib/phase/flatten.js
tkyi: needs to be added here

lib/phase/flatten.js
```diff
@@ -522,6 +537,75 @@ function convertSteps(jobs) {
...
+            stageJobs.forEach(jobName => {
+                if (newJobs[jobName]) {
+                    newJobs[jobName].stage = { name: stageName };
+                    // Handle startFrom
+                    if (
+                        newJobs[jobName].requires &&
+                        newJobs[jobName].requires.includes(`~${getFullStageName({ stageName, type: 'setup' })}`)
+                    ) {
+                        stages[stageName].startFrom = stages[stageName].startFrom
+                            ? stages[stageName].startFrom.push(jobName)
```
sagar1312: `stages[stageName].startFrom.push(jobName)` could throw an error.
First time `startFrom` is set, it is not an array.

lib/phase/flatten.js
sagar1312: Per latest code, we still seem to set `startFrom`

lib/phase/flatten.js
```diff
@@ -522,6 +537,75 @@ function convertSteps(jobs) {
...
+
+            stageJobs.forEach(jobName => {
+                if (newJobs[jobName]) {
+                    newJobs[jobName].stage = { name: stageName };
+                    // Handle startFrom
+                    if (
+                        newJobs[jobName].requires &&
+                        newJobs[jobName].requires.includes(`~${getFullStageName({ stageName, type: 'setup' })}`)
+                    ) {
+                        stages[stageName].startFrom = stages[stageName].startFrom
```
sagar1312: https://github.com/screwdriver-cd/config-parser/pull/143#discussion_r1242307190
Stage would always start from `setup`. Why do we need to set `startFrom` on the stage that points to the job(s) that needs to be triggered after `setup`?

lib/phase/flatten.js
```diff
@@ -522,6 +537,75 @@ function convertSteps(jobs) {
...
+                            ? stages[stageName].startFrom.push(jobName)
+                            : jobName;
+
+                        if (typeof newJobs[jobName].requires === 'string') {
```
sagar1312: Why do we need to exclude `setup` from `requires` of stage job that needs to be triggered after `setup` is completed?
We should retain this information and also persist in DB (`jobs` table) like we do for any other job.

lib/phase/flatten.js
sagar1312: Not yet addressed.
Stage `requires` should always be copied to `setup` job `requires` which would need `setup` to be implicitly created when is it not defined in `screwdriver.yaml`.

lib/phase/flatten.js
sagar1312: Not yet addressed.
We made changes to our initial design to support `setup` and `teardown` so that a stage will always have `setup` and `teardown` even when it is not explicitly defined in `screwdriver.yaml`.
This would make workflow graph consistent irrespective of setup/teardown is configured in the `screwdriver.yaml`.

lib/phase/flatten.js
```diff
@@ -522,6 +537,75 @@ function convertSteps(jobs) {
...
+        Object.keys(stages).forEach(stageName => {
+            const { setup, teardown, jobs: stageJobs } = stages[stageName];
+
+            stageJobs.forEach(jobName => {
+                if (newJobs[jobName]) {
+                    newJobs[jobName].stage = { name: stageName };
+                    // Handle startFrom
+                    if (
+                        newJobs[jobName].requires &&
+                        newJobs[jobName].requires.includes(`~${getFullStageName({ stageName, type: 'setup' })}`)
```
sagar1312: We should not allow the prefix `~` for `setup` in `requires` of a stage job.

Let say a stage `deploy` has a job with requires like `[~stage@deploy:setup, ~jobOutsideStage]`.
This would mean the stage job can be executed after `jobOutsideStage` is completed without waiting for `~stage@deploy:setup`

lib/phase/flatten.js
sagar1312: There is no need to set `{}` in the data-schema. 
Either we would have `setup` coming from `screwdriver.yaml` or implicitly created. Having this set to `{}` would never be utilized.

lib/phase/flatten.js
sagar1312: Same here
There is no need to set `{} `in the data-schema.
Either we would have `teardown` coming from screwdriver.yaml or implicitly created. Having this set to `{}` would never be utilized.

lib/phase/flatten.js
sagar1312: Also `startFrom` is not defined in `stage` domain/DB model. This might be temporary field just utilized in `workflow-parser`. If that is the case, we should keep this localized to `workflow-parser`.

i.e. Instead of `config-parser` checking for the presence of `setup` in `requires` and setting `startFrom` which is then used by `workflow-parser`, we should make `workflow-parser` directly check for the presence of `setup` in `requires`.

lib/phase/permutation.js
sagar1312: Field `stage` stored under `permutations` in DB does not align with the schema
https://github.com/screwdriver-cd/data-schema/pull/491/files#r1312129998

The schema needs to be fixed so that a job can only point to a single stage.

tkyi: https://github.com/screwdriver-cd/data-schema/pull/529

lib/phase/flatten.js
tkyi: updated

lib/phase/flatten.js
tkyi: updated

lib/phase/flatten.js
```diff
@@ -548,6 +548,40 @@ function getFullStageName({ stageName, type }) {
...
+            stageJobNames.push(jobName);
+        }
+    });
+
+    // Get terminal nodes
+    const teardownRequires = stageJobNames.filter(
+        jobName =>
+            !stageJobsRequires.includes(jobName) &&
+            !stageJobsRequires.includes(`~${jobName}`) &&
+            !stageJobsRequires.includes(jobName.split('~')[0])
```
sagar1312: `!stageJobsRequires.includes(jobName.split('~')[0]`
When would we have `~` after job name?

lib/phase/flatten.js
```diff
@@ -3,6 +3,9 @@
 const Hoek = require('@hapi/hoek');
 const clone = require('clone');
 
+const STAGE_PREFIX = 'stage@';
+const DEFAULT_JOB = { image: 'node:18', steps: [{ noop: 'echo noop' }] };
```
sagar1312, line 7: Instead of hardcoding the image to `node:18`, can we use tag? something like `node:current` / `node:latest`
```suggestion
const DEFAULT_JOB = { image: 'node:18', steps: [{ noop: 'echo noop' }] };
```

sagar1312 (changes requested):

lib/phase/flatten.js
tkyi: I think it's better to hardcode so we are sure we support that version.

lib/phase/flatten.js
tkyi: it's if the jobName starts with `~` but the requires list does not have the prefix

sagar1312: Do we allow `~` as a prefix for a job name?
Per schema, looks like we don't allow it.
image

tkyi: removed

sagar1312 (approved):

PR #144, opened 2023-01-20
fix: increase MAX_ENVIRONMENT_VARS from 100 to 120

adong:
## Context

Some users with a large number of environment variables defined in sd.yaml will see issue like 

```
19:48:39 $ echo 'Error: "environment" can only have 100 environment variables defined
19:48:39 Job "publish": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)
19:48:39 "environment" can only have 100 environment variables defined
19:48:39 Job "component": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)
19:48:39 "environment" can only have 100 environment variables defined
19:48:39 Job "owasp": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)'; exit 1
19:48:39 Error: "environment" can only have 100 environment variables defined
19:48:39 Job "publish": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)
19:48:39 "environment" can only have 100 environment variables defined
19:48:39 Job "component": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)
19:48:39 "environment" can only have 100 environment variables defined
19:48:39 Job "owasp": "environment" and "matrix" can only have a combined  maxiumum of 100 environment variables defined (currently 101)```
```

Validator 
image

and build run-time validator
image

## Objective
Increase `MAX_ENVIRONMENT_VARS`, the upper limit of allowed environment variables from `100` to `120`

## License


adong:
Trying to fix tests https://cd.screwdriver.cd/pipelines/18/builds/842485/steps/test
image

tkyi:
Not sure if we should continue to increase the max env var number
I think originally it was a lot lower
Might be good to examine why the customer needs so many

adong:
For users just use screwdriver without using template, 100 is quite a lot. :white_check_mark:

For users who use templates, some [template](https://docs.screwdriver.cd/user-guide/templates.html) itself defines quite a lot. i.e. `maven-open-jdk-rhel7` environment variables. 

Maybe for the flexibility to support multi-platform and multi-version _(I dont know why there’s more than like 50 env being used in this particular template)_

So, the room left for sd users is not much. :x:

----

Also, I’d say, this change actually was just found out by user, and they use latest version tag for `maven-open-jdk-rhel7` template. And `maven-open-jdk-rhel7` recently added some more environment variables, and caused the overflow to the end users’ pipeline.

adong:
This is not a solution, just a patch. If we can find a solution to solve it that would be great, otherwise, this issue will come back sooner or later. 🥲

pritamstyz4ever:
closing due to inactivity

PR #148, opened 2024-02-15
feat(3026): Fill-in stage setup/teardown triggers in job 'requires'

sagar1312:
## Context

Configuring `setup` and `teardown` for a stage is optional. When not specified, `setup` and `teardown` jobs are implicitly created. 

When they are not explicitly configured,  expecting stage `setup` and `teardown` in a job `requires` to define the workflow could be confusing.
Ex:
```
stages:
    canary:
        jobs: [main, publish]

jobs:
    main:
       requires: [~stage@canary:setup]
       .....
       .....
    publish:
       requires: [main]
       .....
       .....
    job-triggered-after-stage:
       requires: [~stage@canary:teardown]
       .....
       .....
       
```

## Objective
Need below changes:
- Allow stage (Ex: `~stage:canary`) to appear as upstream (i.e. in `requires`) for a job outside the stage. This should get translated to `teardown` (Ex: `~stage:canary:teardown`) when the configuration is parsed.
- If a stage job has no or empty `requires`, set stage `setup` (Ex: `~stage:canary:setup`) as upstream

With these changes the screwdriver yaml configuration could be 
```
stages:
    canary:
        jobs: [main, publish]

jobs:
    main:
       requires: []
       .....
       .....
    publish:
       requires: [main]
       .....
       .....
    job-triggered-after-stage:
       requires: [~stage@canary]
       .....
       .....
       
```

## References

https://github.com/screwdriver-cd/screwdriver/issues/3026

## License

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -2,6 +2,7 @@
 
 const Hoek = require('@hapi/hoek');
 const clone = require('clone');
+const { STAGE_TRIGGER } = require('screwdriver-data-schema/config/regex');
 
 const STAGE_PREFIX = 'stage@';
 const DEFAULT_JOB = { image: 'node:18', steps: [{ noop: 'echo noop' }] };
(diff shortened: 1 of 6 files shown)


lib/phase/flatten.js
```diff
@@ -577,6 +578,28 @@ function getTeardownRequires(jobs, stageName) {
     );
 }
 
+/**
+ * Flatten stage setup and teardown into jobs object
```
tkyi: need to update description

sagar1312: Done

tkyi (approved):

PR #149, opened 2024-02-22
feat(3020): add pipeline template support

klu909:
## Context

add pipeline template support
## References

https://github.com/screwdriver-cd/screwdriver/issues/3020
## License

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const RESERVED_PIPELINE_ANNOTATIONS = require('screwdriver-data-schema').config.
 /* eslint-enable max-len */
 
 const phaseValidateStructure = require('./lib/phase/structural');
+const phaseMerge = require('./lib/phase/merge');
 const phaseFlatten = require('./lib/phase/flatten');
 const phaseValidateFunctionality = require('./lib/phase/functional');
 const phaseGeneratePermutations = require('./lib/phase/permutation');
@@ -169,19 +170,25 @@ function verifyStages(stages, jobs) {
 /**
  * Parse the configuration from a screwdriver.yaml
  * @method configParser
- * @param   {Object}               config
- * @param   {String}               config.yaml                Contents of screwdriver.yaml
- * @param   {TemplateFactory}      config.templateFactory     Template Factory to get templates
- * @param   {BuildClusterFactory}  config.buildClusterFactory Build cluster Factory to get build clusters
- * @param   {TriggerFactory}       [config.triggerFactory]    Trigger Factory to find external triggers
- * @param   {Number}               [config.pipelineId]        ID of the current pipeline
- * @param   {Boolean}              [config.notificationsValidationErr]  Throw error when notifications validation fails (default true);
- *                                                                      otherwise return warning
+ * @param   {Object}                          config
+ * @param   {String}                          config.yaml                                 Contents of screwdriver.yaml
+ * @param   {TemplateFactory}                 config.templateFactory                      Template Factory to get templates
+ * @param   {PipelineTemplateVersionFactory}  config.pipelineTemplateVersionFactory       PiplineTemplateVersion Factory to get pipeline templates
+ * @param   {PipelineTemplateTagFactory}      config.pipelineTemplateTagFactory           PipelineTemplateTag Factory to get pipeline templates tag
+ * @param   {PipelineTemplateFactory}         config.pipelineTemplateFactory              PipelineTemplate Factory to get pipeline templates meta
+ * @param   {BuildClusterFactory}             config.buildClusterFactory                  Build cluster Factory to get build clusters
+ * @param   {TriggerFactory}                  [config.triggerFactory]                     Trigger Factory to find external triggers
+ * @param   {Number}                          [config.pipelineId]                         ID of the current pipeline
+ * @param   {Boolean}                         [config.notificationsValidationErr]         Throw error when notifications validation fails (default true);
+ *                                                                                        otherwise return warning
  * @returns {Promise}
  */
 module.exports = function configParser({
     yaml,
     templateFactory,
+    pipelineTemplateVersionFactory,
+    pipelineTemplateTagFactory,
+    pipelineTemplateFactory,
     buildClusterFactory,
     triggerFactory,
     pipelineId,
@@ -192,6 +199,19 @@ module.exports = function configParser({
     // Convert from YAML to JSON
     return (
         parseYaml(yaml)
+            // Merge Pipeline template
+            .then(parsedDoc =>
+                phaseMerge(
+                    parsedDoc,
+                    pipelineTemplateVersionFactory,
+                    pipelineTemplateTagFactory,
+                    pipelineTemplateFactory
+                ).then(({ warnings, mergedDoc }) => {
+                    warnMessages = warnMessages.concat(warnings);
+
+                    return mergedDoc;
+                })
+            )
             // Basic validation
             .then(phaseValidateStructure)
             // Flatten structures
--- a/index.js
+++ b/index.js
@@ -199,6 +199,8 @@ module.exports = function configParser({
     // Convert from YAML to JSON
     return (
         parseYaml(yaml)
+            // Validate user config before mergin with pipeline template
+            .then(doc => phaseValidateStructure(doc, true))
             // Merge Pipeline template
             .then(parsedDoc =>
                 phaseMerge(
@@ -212,7 +214,7 @@ module.exports = function configParser({
                     return mergedDoc;
                 })
             )
-            // Basic validation
+            // Validate merged pipeline config
             .then(phaseValidateStructure)
             // Flatten structures
             .then(parsedDoc => {
--- a/index.js
+++ b/index.js
@@ -270,6 +270,12 @@ module.exports = function configParser({
                     res.stages = stages;
                 }
 
+                const templateVersionId = Hoek.reach(doc, 'templateVersionId');
+
+                if (templateVersionId) {
+                    res.templateVersionId = templateVersionId;
+                }
+
                 return res;
             })
             .catch(err => ({
--- a/lib/phase/merge.js
+++ b/lib/phase/merge.js
@@ -3,6 +3,7 @@
 const Hoek = require('@hapi/hoek');
 const TEMPLATE_NAME_REGEX_WITH_NAMESPACE =
     require('screwdriver-data-schema').config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
+const SCHEMA_CONFIG = require('screwdriver-data-schema').config.base.config;
 
 /**
  * Fetch pipeline template
@@ -79,6 +80,8 @@ function mergeSharedConfig(parsedDoc, pipelineTemplate) {
         newPipeline.shared.image = mergedImage;
     }
 
+    newPipeline.templateVersionId = pipelineTemplate.id;
+
     return newPipeline;
 }
 
@@ -101,6 +104,11 @@ module.exports = async (
 ) => {
     const warnings = [];
     const pipelineTemplateName = Hoek.reach(parsedDoc, 'template');
+
+    // validate screwdriver yaml before merging with pipeline template
+    await SCHEMA_CONFIG.validateAsync(parsedDoc, {
+        abortEarly: false
+    });
 
     if (!pipelineTemplateName) {
         return { mergedDoc: parsedDoc, warnings };
--- a/lib/phase/merge.js
+++ b/lib/phase/merge.js
@@ -1,0 +1,123 @@
+'use strict';
+
+const Hoek = require('@hapi/hoek');
+const TEMPLATE_NAME_REGEX_WITH_NAMESPACE =
+    require('screwdriver-data-schema').config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
+
+/**
+ * Fetch pipeline template
+ * @method fetchPipelineTemplate
+ * @method
+ * @param   {String}                          pipelineTemplateName                        Pipeline template name
+ * @param   {PipelineTemplateVersionFactory}  config.pipelineTemplateVersionFactory       PiplineTemplateVersion Factory to get pipeline templates
+ * @param   {PipelineTemplateTagFactory}      config.pipelineTemplateTagFactory           PipelineTemplateTag Factory to get pipeline templates tag
+ * @param   {PipelineTemplateFactory}         config.pipelineTemplateFactory              PipelineTemplate Factory to get pipeline templates
+ * @return  {Object}                                                                      Pipeline Template
+ */
+async function fetchPipelineTemplate(
+    pipelineTemplateName,
+    pipelineTemplateVersionFactory,
+    pipelineTemplateTagFactory,
+    pipelineTemplateFactory
+) {
+    const [, namespace, name, versionOrTag] = TEMPLATE_NAME_REGEX_WITH_NAMESPACE.exec(pipelineTemplateName);
+    const templateTag = await pipelineTemplateTagFactory.get({
+        name,
+        namespace,
+        tag: versionOrTag
+    });
+    const version = templateTag ? templateTag.version : versionOrTag;
+    const pipelineTemplate = await pipelineTemplateVersionFactory.getWithMetadata(
+        {
+            name,
+            namespace,
+            version
+        },
+        pipelineTemplateFactory
+    );
+
+    return pipelineTemplate;
+}
+
+/**
+ * Merge parsed yaml shared config with pipeline template shared config
+ * Pipeline yaml will take precedence over pipeline template
+ * @method mergeSetting
+ * @method
+ * @param   {Object}        parsedDoc              Parsed screwdriver yaml
+ * @param   {Object}        pipelineTemplate       Pipeline template
+ * @return  {Object}                               Merged yaml settings
+ */
+function mergeSharedConfig(parsedDoc, pipelineTemplate) {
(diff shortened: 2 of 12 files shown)


lib/phase/merge.js
```diff
@@ -0,0 +1,108 @@
...
+    require('screwdriver-data-schema').config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
+
+/**
+ * Fetch pipeline template
+ * @method fetchPipelineTemplate
+ * @method
+ * @param   {String}                          pipelineTemplateName                        Pipeline template name
+ * @param   {PipelineTemplateVersionFactory}  config.pipelineTemplateVersionFactory       PiplineTemplateVersion Factory to get pipeline templates
+ * @param   {PipelineTemplateTagFactory}      config.pipelineTemplateTagFactory           PipelineTemplateTag Factory to get pipeline templates tag
+ * @param   {TemplateMetaFactory}             config.templateMetaFactory                  TemplateMeta Factory to get pipeline templates
```
pritamstyz4ever: we do not use `TemplateMetaFactory`, we need to use `pipelineTemplateFactory`

pritamstyz4ever (changes requested):

lib/phase/merge.js
klu909: https://github.com/screwdriver-cd/models/blob/41293848c21da3fc419993c2188f969fdde44d39/lib/pipelineTemplateVersionFactory.js#L217

klu909: `getWithMetadata` uses `TemplateMetaFactory`

pritamstyz4ever: that is the base class, you need to pass `pipelineTemplateFactory` instead

index.js
```diff
@@ -169,19 +170,25 @@ function verifyStages(stages, jobs) {
...
- * @param   {TemplateFactory}      config.templateFactory     Template Factory to get templates
- * @param   {BuildClusterFactory}  config.buildClusterFactory Build cluster Factory to get build clusters
- * @param   {TriggerFactory}       [config.triggerFactory]    Trigger Factory to find external triggers
- * @param   {Number}               [config.pipelineId]        ID of the current pipeline
- * @param   {Boolean}              [config.notificationsValidationErr]  Throw error when notifications validation fails (default true);
- *                                                                      otherwise return warning
+ * @param   {Object}                          config
+ * @param   {String}                          config.yaml                                 Contents of screwdriver.yaml
+ * @param   {TemplateFactory}                 config.templateFactory                      Template Factory to get templates
+ * @param   {PipelineTemplateVersionFactory}  config.pipelineTemplateVersionFactory       PiplineTemplateVersion Factory to get pipeline templates
```
pritamstyz4ever, line 176: PipelineTemplateVersion

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+        SD_PIPELINE_TEMPLATE_VERSION: pipelineTemplate.version
+    });
+
+    templateSharedConfig.environment = Hoek.merge(templateSharedConfig.environment, yamlSharedConfig.environment || {});
+    templateSharedConfig.settings = Hoek.merge(templateSharedConfig.settings || {}, yamlSharedConfig.settings || {});
+
+    if (yamlSharedConfig.image || templateSharedConfig.image) {
+        templateSharedConfig.image = yamlSharedConfig.image || templateSharedConfig.image;
+    }
+
```
pritamstyz4ever: It isn't easy to understand the intention of this logic. If the problem is that yamlSharedConfig.image should not overwrite templateSharedConfig.image when templateSharedConfig.image is truthy, then the code can be rewritten as follows:
`
if (yamlSharedConfig.image && !templateSharedConfig.image) {
    templateSharedConfig.image = yamlSharedConfig.image;
}
`

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Merge parsed yaml setting with pipeline template setting
+ * Pipeline yaml will take precedence over pipeline template
+ * @method mergeSetting
+ * @method
+ * @param   {Object}        parsedDoc              Parsed screwdriver yaml
+ * @param   {Object}        pipelineTemplate       Pipeline template
+ * @return  {Object}                               Merged yaml settings
+ */
+function mergeSetting(parsedDoc, pipelineTemplate) {
```
pritamstyz4ever:
```suggestion
function mergeSetting(parsedDoc, pipelineTemplate) {
    const newPipeline = { ...pipelineTemplate.config };
    const { shared: yamlSharedConfig } = parsedDoc;
    const { shared: templateSharedConfig } = newPipeline;

    const mergedEnvironment = {
        SD_PIPELINE_TEMPLATE_FULLNAME: parsedDoc.template,
        SD_PIPELINE_TEMPLATE_NAME: pipelineTemplate.name,
        SD_PIPELINE_TEMPLATE_NAMESPACE: pipelineTemplate.namespace,
        SD_PIPELINE_TEMPLATE_VERSION: pipelineTemplate.version,
        ...templateSharedConfig.environment,
        ...yamlSharedConfig.environment
    };

    const mergedSettings = {
        ...templateSharedConfig.settings,
        ...yamlSharedConfig.settings
    };

    const mergedImage = yamlSharedConfig.image || templateSharedConfig.image;

    newPipeline.shared = {
        ...templateSharedConfig,
        environment: mergedEnvironment,
        settings: mergedSettings,
        image: mergedImage
    };

    return newPipeline;
}
```

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+    parsedDoc,
+    pipelineTemplateVersionFactory,
+    pipelineTemplateTagFactory,
+    pipelineTemplateFactory
+) => {
+    const warnings = [];
+    const pipelineTemplateName = Hoek.reach(parsedDoc, 'template');
+
+    if (!pipelineTemplateName) {
+        return Promise.resolve({ mergedDoc: parsedDoc, warnings });
```
pritamstyz4ever: you don't need to use Promise.resolve({}) in an async function. The async keyword automatically wraps the return value of the function in a Promise. If you return an object, it will be equivalent to returning Promise.resolve({}).

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+    if (yamlSharedConfig.image || templateSharedConfig.image) {
+        templateSharedConfig.image = yamlSharedConfig.image || templateSharedConfig.image;
+    }
+
+    newPipeline.shared = templateSharedConfig;
+
+    return newPipeline;
+}
+
+/**
```
pritamstyz4ever, line 87:
```suggestion
/**
 * Merge pipeline template with screwdriver yaml
 * @method mergePipelineTemplate
 * @param   {Object}                          parsedDoc                                   Document that went through structural parsing
 * @param   {PipelineTemplateVersionFactory}  pipelineTemplateVersionFactory              PipelineTemplateVersion Factory to get pipeline templates
 * @param   {PipelineTemplateTagFactory}      pipelineTemplateTagFactory                  PipelineTemplateTag Factory to get pipeline templates tag
 * @param   {PipelineTemplateFactory}         pipelineTemplateFactory                     PipelineTemplate Factory to get pipeline templates
 * @returns {Promise}                                                                     Merged document and warnings
 */
async function mergePipelineTemplate(
    parsedDoc,
    pipelineTemplateVersionFactory,
    pipelineTemplateTagFactory,
    pipelineTemplateFactory
) {
    const warnings = [];
    const pipelineTemplateName = Hoek.reach(parsedDoc, 'template');

    if (!pipelineTemplateName) {
        return { mergedDoc: parsedDoc, warnings };
    }

    const pipelineTemplate = await fetchPipelineTemplate(
        pipelineTemplateName,
        pipelineTemplateVersionFactory,
        pipelineTemplateTagFactory,
        pipelineTemplateFactory
    );

    if (!pipelineTemplate) {
        throw new Error(`Pipeline template ${pipelineTemplateName} does not exist`);
    }

    const mergedDoc = mergeSetting(parsedDoc, pipelineTemplate);

    return { mergedDoc, warnings };
}

module.exports = mergePipelineTemplate;
```

test/index.test.js
```diff
@@ -95,20 +101,11 @@ describe('config parser', () => {
...
                 assert.strictEqual(data.jobs.main[0].commands[0].name, 'config-parse-error');
-                assert.match(data.jobs.main[0].commands[0].command, /ValidationError:.*"jobs" is required/);
-                assert.match(data.errors[0], /ValidationError:.*"jobs" is required/);
+                assert.match(data.errors[0], 'ValidationError: "jobs" is required');
             }));
     });
 
     describe('structure validation', () => {
-        describe('overall config', () => {
-            it('returns an error if missing jobs', () =>
```
pritamstyz4ever: I think we need to keep this test and modify it to say `returns an error if missing jobs and no template specified`

pritamstyz4ever (changes requested):

index.js
```diff
@@ -192,6 +199,19 @@ module.exports = function configParser({
...
+                phaseMerge(
+                    parsedDoc,
+                    pipelineTemplateVersionFactory,
+                    pipelineTemplateTagFactory,
+                    pipelineTemplateFactory
+                ).then(({ warnings, mergedDoc }) => {
+                    warnMessages = warnMessages.concat(warnings);
+
+                    return mergedDoc;
+                })
```
sagar1312, line 215: Shouldn't we merge pipeline template config into pipeline config after structural validation?
If we merge first, how are we enforcing the schema restrictions introduced in this PR https://github.com/screwdriver-cd/data-schema/pull/551

We may have to perform structural validation both before and after merging with slight variations in the rules.

lib/phase/merge.js
sagar1312: The scope of this function is wider than just merging the `setting`. Can we rename accordingly?

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+ * @param   {Object}        pipelineTemplate       Pipeline template
+ * @return  {Object}                               Merged yaml settings
+ */
+function mergeSetting(parsedDoc, pipelineTemplate) {
+    const newPipeline = pipelineTemplate.config;
+    const yamlSharedConfig = parsedDoc.shared || {};
+    const templateSharedConfig = newPipeline.shared || {};
+
+    templateSharedConfig.environment = Hoek.merge(templateSharedConfig.environment || {}, {
+        SD_PIPELINE_TEMPLATE_FULLNAME: parsedDoc.template,
```
sagar1312: Shouldn't the assigned value exclude version/tag?
Refer: Job template https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/flatten.js#L342

```suggestion
        SD_PIPELINE_TEMPLATE_FULLNAME: `${pipelineTemplate.namespace}/${pipelineTemplate.name}`,
```

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    templateSharedConfig.environment = Hoek.merge(templateSharedConfig.environment, yamlSharedConfig.environment || {});
+    templateSharedConfig.settings = Hoek.merge(templateSharedConfig.settings || {}, yamlSharedConfig.settings || {});
+
+    if (yamlSharedConfig.image || templateSharedConfig.image) {
+        templateSharedConfig.image = yamlSharedConfig.image || templateSharedConfig.image;
+    }
+
+    newPipeline.shared = templateSharedConfig;
```
sagar1312: Per design, we also need to set `templateVersionId`.
Also refer job template: https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/flatten.js#L400

```suggestion
    newPipeline.shared = templateSharedConfig;
    
    newPipeline.templateVersionId = pipelineTemplate.id: 
```

lib/phase/merge.js
```diff
@@ -0,0 +1,113 @@
...
+ * @method mergeSetting
+ * @method
+ * @param   {Object}        parsedDoc              Parsed screwdriver yaml
+ * @param   {Object}        pipelineTemplate       Pipeline template
+ * @return  {Object}                               Merged yaml settings
+ */
+function mergeSetting(parsedDoc, pipelineTemplate) {
+    const newPipeline = pipelineTemplate.config;
+    const yamlSharedConfig = parsedDoc.shared || {};
+    const templateSharedConfig = newPipeline.shared || {};
```
sagar1312, line 54: Consider renaming
```suggestion
    const newSharedConfig = newPipeline.shared || {};
```

lib/phase/structural.js
```diff
@@ -12,6 +12,10 @@ const SCHEMA_CONFIG = require('screwdriver-data-schema').config.base.config;
 function checkAdditionalRules(data) {
     const { jobs } = data;
 
+    if (!jobs) {
+        throw new Error('Pipeline has no jobs');
+    }
```
sagar1312: Current implementation performs structural validation after merging pipeline template config into to pipeline config.
Why/when would `jobs` be `undefined`?

test/index.test.js
```diff
@@ -556,6 +553,60 @@ describe('config parser', () => {
...
+                    templateFactory: templateFactoryMock,
+                    triggerFactory,
+                    pipelineTemplateTagFactory: pipelineTemplateTagFactoryMock,
+                    pipelineTemplateVersionFactory: pipelineTemplateVersionFactoryMock
+                }).then(data => {
+                    assert.deepEqual(data, JSON.parse(loadData('pipeline-template-with-shared-setting-result.json')));
+                });
+            });
+
+            it('returns error if pipeline template does not exist', () => {
```
sagar1312, line 615: Add tests to cover scenario when forbidden fields (Ex: `jobs`, `parameters`, `stages`) haven been specified in the pipeline configuration which uses pipeline template

sagar1312 (changes requested):

index.js
klu909: will think more on this, but idea is to verify that merged yaml passes validation

lib/phase/structural.js
klu909: that that check happens even when you are not using pipeline template

test/index.test.js
klu909: ~that's out of scope, adding it to data-schema is enough~, added one, but it will need to be changed as we relax the limitation

index.js
klu909: what does that suppose to mean? what need to be change here?

lib/phase/structural.js
sagar1312: Yes, in that case it should be caught when we validate the config against the schema.
Why do we need to explicitly check?

lib/phase/merge.js
klu909: ` const { shared: yamlSharedConfig } = parsedDoc;` has to stay as `    const yamlSharedConfig = parsedDoc.shared || {};`. It would throw error when you try to do yamlSharedConfig, unless we want to add `if` statement to guard that

test/index.test.js
klu909: yeah should add it back, i deleted it when i placed `merge` after `structure validation`. now `merge` happens before `structure validation` so it should pass now

lib/phase/merge.js
klu909: I would agree, but we need to alter `pipeline` table to add this. or we can add it to environment variable

klu909: thoughts? cc @pritamstyz4ever

klu909:
https://github.com/screwdriver-cd/data-schema/pull/554 pr to add column `templateVersionId`, test will fail until we merge that pr

lib/phase/merge.js
klu909: also need additional change in model to pick up this value

index.js
klu909: added another check right before merging

klu909:
will need additional pr on models to pick up `templateVersionId` and update pipeline with that value, will do that after we close out all existing pr

index.js
sagar1312: Let us move the logic to validate out of `merge.js`. It will be good not to overload a module to handle multiple things (validating, merging, etc).

We can use the existing `structural.js` module to perform validation before/after merging. We can modify it to take a boolean flag `skipAdditionalRules`.

sagar1312 (changes requested):

index.js
pritamstyz4ever: you have a typo https://github.com/screwdriver-cd/config-parser/pull/149/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R176

test/index.test.js
```diff
@@ -95,8 +101,7 @@ describe('config parser', () => {
                 assert.deepEqual(data.jobs.main[0].secrets, []);
                 assert.deepEqual(data.jobs.main[0].environment, {});
                 assert.strictEqual(data.jobs.main[0].commands[0].name, 'config-parse-error');
-                assert.match(data.jobs.main[0].commands[0].command, /ValidationError:.*"jobs" is required/);
```
pritamstyz4ever, line 98: is this assertion not needed?

pritamstyz4ever (commented):
Please fix tests

pritamstyz4ever (approved):

test/index.test.js
klu909: emm, it's a dup with `data.errors[0]`, not really needed

sagar1312 (approved):
Looks good. We need to bump up the major version because of the breaking change in one of the dependency.
https://github.com/screwdriver-cd/data-schema/pull/554

PR #155, opened 2024-03-18
feat(3059): Add pipeline level customization for pipeline template

klu909:
## Context

Add pipeline level customization for pipeline template
## Objective
Add pipeline level customization for pipeline template

## References
https://github.com/screwdriver-cd/screwdriver/issues/3058

## License


lib/phase/merge.js
```diff
@@ -79,6 +79,32 @@ function mergeSharedConfig(parsedDoc, pipelineTemplate) {
         newPipeline.shared.image = mergedImage;
     }
 
+    if (parsedDoc.cache) {
```
sagar1312, line 82: We may have to concatenate the cache paths from the pipeline configuration with cache paths inherited from the pipeline template for all the scopes (`pipeline`, `event`, `jobs`).
If we do high level replace, the cache paths configured for pipeline template jobs will be lost.

lib/phase/merge.js
```diff
@@ -79,6 +79,32 @@ function mergeSharedConfig(parsedDoc, pipelineTemplate) {
...
+        newPipeline.cache = {
+            ...newPipeline.cache,
+            ...parsedDoc.cache
+        };
+    }
+
+    newPipeline.subscribe = {
+        ...newPipeline.subscribe,
+        ...parsedDoc.subscribe
+    };
```
sagar1312, line 92: Same here. Instead of high level replace (i.e. pipeline `subscribe.scmUrls` completely replacing pipeline template `subscribe.scmUrls` ), should we append the pipeline `subscribe.scmUrls` to pipeline template `subscribe.scmUrls`?

test/data/pipeline-template-with-pipeline-setting.json
```diff
@@ -0,0 +1,39 @@
+{
+    "id":111,
+    "version": "1.0.0",
+    "templateId": 1234,
+    "description": "Template for building a NodeJS module\nInstalls dependencies and runs tests\n",
+    "config": {
```
sagar1312: Please add tests to cover below scenarios

1. The fields exists in the pipeline using the template but not in the template
2. The fields does not exists in the pipeline using the template but exists in the template
3. The fields exists both in the pipeline using the template and in the template. We need to cover sub-scenarios like adding new entry vs replacing an entry vs appending to an entry, etc

Looks like the current test just covers `1`.

sagar1312 (changes requested):

lib/phase/merge.js
klu909: I think this one should be a replace, is there a use case for multiple cache paths?

lib/phase/merge.js
klu909: this honestly it could go both ways, what I was thinking is that user may or maybe not want subscribe defined by pipeline template, if they want to retain that same scmurl while adding new ones, then they should add both to their yaml. otherwise how do you handle the case where user that just want to replace it

PR #157, opened 2024-04-23
feat(3068): flatten pipeline template

klu909:
## Context

flatten pipeline template by eliminating shared settings 
## Objective
remove shared settings from merging logic and flatten shared into jobs

## References
https://github.com/screwdriver-cd/screwdriver/issues/3068

## License

### DIFF
--- a/index.js
+++ b/index.js
@@ -319,7 +319,7 @@ function parsePipelineYaml({
  * @param   {String}          config.yaml           Pipeline Template
  * @return  {Object}                                Pipeline Template
  */
-async function parsePipelineTemplate(yaml) {
+async function parsePipelineTemplate({ yaml }) {
     const configToValidate = await parseYaml(yaml);
     const pipelineTemplate = await SCHEMA_PIPELINE_TEMPLATE.validateAsync(configToValidate, {
         abortEarly: false
@@ -327,7 +327,7 @@ async function parsePipelineTemplate(yaml) {
     const pipelineTemplateConfig = pipelineTemplate.config;
 
     // flatten pipeline template shared setting into jobs
-    pipelineTemplateConfig.job = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.job);
+    pipelineTemplateConfig.jobs = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.jobs);
     delete pipelineTemplateConfig.shared;
 
     return pipelineTemplate;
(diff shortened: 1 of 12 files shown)


lib/phase/merge.js
```diff
@@ -161,36 +161,28 @@ function mergePipelineLevelConfig(parsedDoc, pipelineTemplate) {
...
-    if (mergedImage) {
-        newPipeline.shared.image = mergedImage;
+    if (yamlSharedConfig.image) {
+        newPipeline.shared.image = yamlSharedConfig.image;
     }
 
     newPipeline.templateVersionId = pipelineTemplate.id;
 
+    console.log(newPipeline);
+
```
tkyi:
```suggestion
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-config-parser",
-  "version": "8.0.0",
+  "version": "10.0.0",
```
VonnyJap, line 3:
```suggestion
  "version": "9.0.0",
```

index.js
```diff
@@ -308,4 +310,30 @@ module.exports = function configParser({
...
+        abortEarly: false
+    });
+    const pipelineTemplateConfig = pipelineTemplate.config;
+
+    // flatten pipeline template shared setting into jobs
+    pipelineTemplateConfig.jobs = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.jobs);
+    delete pipelineTemplateConfig.shared;
+
+    return pipelineTemplate;
+}
```
VonnyJap, line 334: where would this function be used? does this function need to be incorporated as well into `screwdriver-models` package?

package.json
klu909: we are already at major version 9, it just didn't get changed last time. https://www.npmjs.com/package/screwdriver-config-parser

index.js
klu909: validator will call this, then screwdriver api calls validator https://github.com/screwdriver-cd/screwdriver/blob/11813f8ac54a3b3bed4e4298ce7ba754c7b7d992/plugins/pipelines/templates/create.js#L22

VonnyJap (approved):

VonnyJap:
@klu909 - PR is approved, please help to propagate the change to other affected modules. 
And another thing, do we need to remove shared from [here](https://github.com/screwdriver-cd/data-schema/blob/54a8ab5985dbe76b456f07e1dc2e75dabbfc1dbb/config/pipelineTemplate.js#L13)?

klu909:
> here

don't think we can, initial user input will have `shared`, we just don't store them

PR #158, opened 2024-05-07
fix(3105): Handle mergeSharedSteps annotation for pipeline template

tkyi:
## Context

Would be nice if the pipeline template could use job templates and `mergeSharedSteps` annotation.

## Objective

This PR allows pipeline template to use job template and `mergeSharedSteps` annotation.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/3105

## License

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,11 @@ const SCHEMA_PIPELINE_TEMPLATE = require('screwdriver-data-schema').config.pipel
 
 const phaseValidateStructure = require('./lib/phase/structural');
 const phaseMerge = require('./lib/phase/merge');
-const { flattenSharedIntoJobs, flattenPhase: phaseFlatten } = require('./lib/phase/flatten');
+const {
+    flattenSharedIntoJobs,
+    handleMergeSharedStepsAnnotation,
+    flattenPhase: phaseFlatten
+} = require('./lib/phase/flatten');
 const phaseValidateFunctionality = require('./lib/phase/functional');
 const phaseGeneratePermutations = require('./lib/phase/permutation');
 
@@ -315,11 +319,10 @@ function parsePipelineYaml({
  * Parses pipeline template configuration
  * @method parsePipelineTemplate
  * @param   {Object}          config
- * @param   {String}          config.yaml             Pipeline Template
- * @param   {TemplateFactory} config.templateFactory  Template Factory to get templates
- * @return  {Object}                                  Pipeline Template
+ * @param   {String}          config.yaml           Pipeline Template
+ * @return  {Object}                                Pipeline Template
  */
-async function parsePipelineTemplate({ yaml, templateFactory }) {
+async function parsePipelineTemplate({ yaml }) {
     const configToValidate = await parseYaml(yaml);
     const pipelineTemplate = await SCHEMA_PIPELINE_TEMPLATE.validateAsync(configToValidate, {
         abortEarly: false
@@ -329,9 +332,17 @@ async function parsePipelineTemplate({ yaml, templateFactory }) {
     // flatten pipeline template shared setting into jobs
     pipelineTemplateConfig.jobs = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.jobs);
 
-    const { flattenedDoc } = await phaseFlatten(pipelineTemplateConfig, templateFactory);
+    const mergedJobs = {};
 
-    pipelineTemplateConfig.jobs = flattenedDoc.jobs;
+    // Handle mergeSharedSteps annotation
+    Object.keys(pipelineTemplateConfig.jobs).forEach(j => {
+        mergedJobs[j] = handleMergeSharedStepsAnnotation({
+            sharedConfig: pipelineTemplateConfig.shared,
+            jobConfig: pipelineTemplateConfig.jobs[j]
+        });
+    });
+    delete pipelineTemplateConfig.shared;
+    pipelineTemplateConfig.jobs = mergedJobs;
 
     return pipelineTemplate;
 }
--- a/index.js
+++ b/index.js
@@ -12,11 +12,7 @@ const SCHEMA_PIPELINE_TEMPLATE = require('screwdriver-data-schema').config.pipel
 
 const phaseValidateStructure = require('./lib/phase/structural');
 const phaseMerge = require('./lib/phase/merge');
-const {
-    flattenSharedIntoJobs,
-    handleMergeSharedStepsAnnotation,
-    flattenPhase: phaseFlatten
-} = require('./lib/phase/flatten');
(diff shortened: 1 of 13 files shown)


lib/phase/flatten.js
```diff
@@ -306,6 +306,46 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+                template.annotations['screwdriver.cd/mergeSharedSteps']))
+    ) {
+        // convert steps from jobConfig from array to object for faster lookup
+        const jobSteps = jobConfig.steps.reduce((obj, item) => {
+            const key = Object.keys(item)[0];
+
+            obj[key] = item[key];
+
+            return obj;
+        }, {});
```
VonnyJap, line 339: smart implementation

lib/phase/flatten.js
```diff
@@ -306,6 +306,46 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+ * @return {Object}                            Merged job config
+ */
+function handleMergeSharedStepsAnnotation({ sharedConfig, jobConfig, template }) {
+    const sharedSteps = sharedConfig.steps;
+    const newJobConfig = jobConfig;
+
+    // merge shared steps into oldJob
+    if (
+        sharedSteps &&
+        ((jobConfig.annotations && jobConfig.annotations['screwdriver.cd/mergeSharedSteps']) ||
```
VonnyJap, line 323: should we also check that jobConfig is using a template? I think that is one of the conditions to perform merging.

index.js
```diff
@@ -328,7 +331,18 @@ async function parsePipelineTemplate({ yaml }) {
...
     pipelineTemplateConfig.jobs = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.jobs);
+
+    const mergedJobs = {};
+
+    // Handle mergeSharedSteps annotation
+    Object.keys(pipelineTemplateConfig.jobs).forEach(j => {
+        mergedJobs[j] = handleMergeSharedStepsAnnotation({
+            sharedConfig: pipelineTemplateConfig.shared,
+            jobConfig: pipelineTemplateConfig.jobs[j]
+        });
```
VonnyJap, line 308: handleMergeSharedStepsAnnotation accepts 3 arguments, only 2 passed here?

lib/phase/flatten.js
```diff
@@ -306,6 +306,46 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+/**
+ * Merge shared steps into job when mergeSharedSteps annotation is set to true
+ * @param  {Object} sharedConfig               Shared config
+ * @param  {Object} jobConfig                  Job config
+ * @param  {Object} [template]                 Template
+ * @return {Object}                            Merged job config
+ */
+function handleMergeSharedStepsAnnotation({ sharedConfig, jobConfig, template }) {
+    const sharedSteps = sharedConfig.steps;
+    const newJobConfig = jobConfig;
```
VonnyJap, line 320: for jobs that use a template, at this point, the steps need be expanded and treated as the steps of this job and subsequently merged with the shared steps if any. I think the current implementation does not handle the expansion of the steps from job template?

index.js
```diff
@@ -332,17 +329,9 @@ async function parsePipelineTemplate({ yaml }) {
...
 
-    // Handle mergeSharedSteps annotation
-    Object.keys(pipelineTemplateConfig.jobs).forEach(j => {
-        mergedJobs[j] = handleMergeSharedStepsAnnotation({
-            sharedConfig: pipelineTemplateConfig.shared,
-            jobConfig: pipelineTemplateConfig.jobs[j]
-        });
-    });
-    delete pipelineTemplateConfig.shared;
-    pipelineTemplateConfig.jobs = mergedJobs;
```
VonnyJap: so we no longer need to use `handleMergeSharedStepsAnnotation`? if so can you remove this function from the related file? and what is the reason of removing `delete pipelineTemplateConfig.shared;`?

index.js
```diff
@@ -319,10 +315,11 @@ function parsePipelineYaml({
...
  * @method parsePipelineTemplate
  * @param   {Object}          config
- * @param   {String}          config.yaml           Pipeline Template
- * @return  {Object}                                Pipeline Template
+ * @param   {String}          config.yaml             Pipeline Template
+ * @param   {TemplateFactory} config.templateFactory  Template Factory to get templates
+ * @return  {Object}                                  Pipeline Template
  */
-async function parsePipelineTemplate({ yaml }) {
+async function parsePipelineTemplate({ yaml, templateFactory }) {
```
VonnyJap: since this is a breaking change, can you also reflect the change to `template-validator` and `screwdriver` repo?

index.js
tkyi: third one is not required

VonnyJap: when will 3rd one ever required then? 3rd one does not have a default value, so it will always become undefined when not supplied during function call

lib/phase/flatten.js
```diff
@@ -307,6 +307,46 @@ function flattenSharedIntoJobs(shared, jobs) {
...
+    // merge shared steps into oldJob
+    if (
+        sharedSteps &&
+        ((jobConfig.annotations && jobConfig.annotations['screwdriver.cd/mergeSharedSteps']) ||
+            ((jobConfig.annotations === undefined ||
+                jobConfig.annotations['screwdriver.cd/mergeSharedSteps'] === undefined) &&
+                template &&
+                template.annotations &&
+                template.annotations['screwdriver.cd/mergeSharedSteps']))
+    ) {
```
VonnyJap, line 331: if a job is not using a template, will it go into this condition? 
Does the second example in [this merging-shared-steps document](https://docs.screwdriver.cd/user-guide/templates/job-templates#merging-with-shared-steps) even valid?

index.js
```diff
@@ -294,7 +297,18 @@ async function parsePipelineTemplate({ yaml }) {
 
     // flatten pipeline template shared setting into jobs
     pipelineTemplateConfig.jobs = flattenSharedIntoJobs(pipelineTemplateConfig.shared, pipelineTemplateConfig.jobs);
```
VonnyJap, line 299: at this point, actually shared steps have been merged into jobs regardless whether or not `screwdriver.cd/mergeSharedSteps` has been set to true, so in a later check of `handleMergeSharedStepsAnnotation`, even though it never go into [the if condition](https://github.com/screwdriver-cd/config-parser/blob/mergeSharedSteps/lib/phase/flatten.js#L322-L330), the shared steps will still be carried along.

lib/phase/flatten.js
tkyi: no it's missing a template we talked about this

index.js
tkyi: ?

VonnyJap (approved):

PR #159, opened 2024-05-08
feat(3102): parser to return an error if a job within a stage is triggered by jobs that are not within the same stage

VonnyJap:
## Context

Jobs within a stage can only be triggered by either a setup job within the same stage or by other jobs also located within that stage

## Objective

This change enforces the restriction described above. Consequently, it will generate an error message if violated.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3102

## License

### DIFF
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -302,8 +302,8 @@ function validateStages(stages, jobs) {
         }
         stage.jobs.forEach(jobName => {
             if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
-                throw new YamlParser.YAMLException(
-                    `${jobName} job has invalid requires: ${jobs[jobName].requires}, triggers must be in the same stage`
+                errors.push(
+                    `${jobName} job has invalid requires: ${jobs[jobName].requires}. This job belongs to the stage {stageName} and is allowed only to have other jobs of the same stage in requires.`
                 );
             }
         });
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -296,6 +296,7 @@ function validateStages(stages, jobs) {
     Object.entries(stages).forEach(([stageName, stage]) => {
         const jobsInStage = [...stage.jobs];
         // Check if setup job is not in the list, then add it
+
         if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
             jobsInStage.push(`stage@${stageName}:setup`);
         }
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -307,7 +307,7 @@ function validateStages(stages, jobs) {
             if (jobs[jobName] && jobs[jobName].requires) {
                 if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
                     errors.push(
-                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers have to be jobs in ${stageName} stage.`
+                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers must be jobs from  ${stageName} stage.`
                     );
                 }
             }
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -273,8 +273,6 @@ function validateStages(stages, jobs) {
         return errors;
     }
 
-    console.log('stages:', stages);
-    console.log('jobs:', jobs);
     // Get list of job names in jobs
     const jobNames = Object.keys(jobs);
 
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -283,7 +283,7 @@ function validateStages(stages, jobs) {
     const duplicateJobsInStage = stageJobNames.filter((item, index) => stageJobNames.indexOf(item) !== index);
 
     if (duplicateJobsInStage.length > 0) {
-        throw new YamlParser.YAMLException(`Cannot have duplicate job in multiple stages: ${duplicateJobsInStage}`);
+        errors.push(`Cannot have duplicate job in multiple stages: ${duplicateJobsInStage}`);
     }
 
     // If job name does not exist, throw error
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -278,7 +278,7 @@ function verifyStages(stages, jobs) {
     Object.values(stages).forEach(stage => {
         stageJobNames = stageJobNames.concat(stage.jobs);
     });
-    console.log(stageJobNames);
+
     // If job name is repeated in stages, throw error
     const duplicateJobsInStage = stageJobNames.filter((item, index) => stageJobNames.indexOf(item) !== index);
 
@@ -297,7 +297,7 @@ function verifyStages(stages, jobs) {
         if (!stages[stageName].jobs || stages[stageName].jobs.length === 0) {
             throw new YamlParser.YAMLException(`Cannot have stage ${stageName} without jobs`);
         }
-        const jobsInStage = stages[stageName].jobs;
+        const jobsInStage = [...stages[stageName].jobs];
         // Check if setup job is not in the list, then add it
 
         if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -302,6 +302,7 @@ function validateStages(stages, jobs) {
         if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
             jobsInStage.push(`stage@${stageName}:setup`);
         }
+        // Make sure requires jobs are all stage jobs
         stage.jobs.forEach(jobName => {
             if (jobs[jobName] && jobs[jobName].requires) {
                 if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -299,7 +299,6 @@ function validateStages(stages, jobs) {
     Object.entries(stages).forEach(([stageName, stage]) => {
         const jobsInStage = [...stage.jobs];
         // Check if setup job is not in the list, then add it
-
         if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
             jobsInStage.push(`stage@${stageName}:setup`);
         }
--- a/lib/phase/functional.js
+++ b/lib/phase/functional.js
@@ -298,6 +298,7 @@ function validateStages(stages, jobs) {
 
     Object.entries(stages).forEach(([stageName, stage]) => {
         const jobsInStage = [...stage.jobs];
+
         // Check if setup job is not in the list, then add it
         if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
             jobsInStage.push(`stage@${stageName}:setup`);
@@ -307,7 +308,7 @@ function validateStages(stages, jobs) {
             if (jobs[jobName] && jobs[jobName].requires) {
                 if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
                     errors.push(
-                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers must be jobs from  ${stageName} stage.`
+                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers must be jobs from ${stageName} stage.`
                     );
                 }
             }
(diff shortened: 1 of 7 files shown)


lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+    const nonexistentJobsInStage = stageJobNames.filter(jobName => !jobNames.includes(jobName));
+
+    if (nonexistentJobsInStage.length > 0) {
+        throw new YamlParser.YAMLException(`Cannot have nonexistent job in stages: ${nonexistentJobsInStage}`);
+    }
+
+    Object.keys(stages).forEach(stageName => {
+        if (!stages[stageName].jobs || stages[stageName].jobs.length === 0) {
+            throw new YamlParser.YAMLException(`Cannot have stage ${stageName} without jobs`);
+        }
```
sagar1312: This would have already been captured during structure validation based on the schema
https://github.com/screwdriver-cd/data-schema/blob/master/config/base.js#L44
```suggestion
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+/**
+ * Check there are no duplicate jobs in stages and all jobs listed exist
+ * @method verifyStages
+ * @param  {Object} stages Stages
+ * @param  {Object} jobs   Jobs
+ */
+function verifyStages(stages, jobs) {
+    if (!stages) {
+        return;
+    }
```
sagar1312: const errors = [];

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+
+    Object.values(stages).forEach(stage => {
+        stageJobNames = stageJobNames.concat(stage.jobs);
+    });
+
+    // If job name is repeated in stages, throw error
+    const duplicateJobsInStage = stageJobNames.filter((item, index) => stageJobNames.indexOf(item) !== index);
+
+    if (duplicateJobsInStage.length > 0) {
+        throw new YamlParser.YAMLException(`Cannot have duplicate job in multiple stages: ${duplicateJobsInStage}`);
```
sagar1312:
```suggestion
        errors.push(`Cannot have duplicate job in multiple stages: ${duplicateJobsInStage}`);
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+    }
+
+    // If job name does not exist, throw error
+    const nonexistentJobsInStage = stageJobNames.filter(jobName => !jobNames.includes(jobName));
+
+    if (nonexistentJobsInStage.length > 0) {
+        throw new YamlParser.YAMLException(`Cannot have nonexistent job in stages: ${nonexistentJobsInStage}`);
```
sagar1312:
```suggestion
        errors.push(`Cannot have nonexistent job in stages: ${nonexistentJobsInStage}`);
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+    }
+
+    Object.keys(stages).forEach(stageName => {
```
sagar1312: Use `Object.entries` instead of `Object.keys(stages)`, to avoid doing look up to get stage suing stage name.
```suggestion
    Object.entries(stages).forEach(([stageName, stage])=> {
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+        // Check if setup job is not in the list, then add it
+
+        if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
+            jobsInStage.push(`stage@${stageName}:setup`);
+        }
+        stages[stageName].jobs.forEach(jobName => {
+            if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
+                throw new YamlParser.YAMLException(
+                    `${jobName} job has invalid requires: ${jobs[jobName].requires}, triggers must be in the same stage`
+                );
```
sagar1312:
```suggestion
                errors.push(
                    `${jobName} job has invalid requires: ${jobs[jobName].requires}. This job belongs to the stage {stageName} and is allowed only to have other jobs of the same stage in requires.`
                );
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
...
+            }
+        });
+    });
+}
```
sagar1312:
```suggestion
    });
    
    return errors;
}
```

lib/phase/functional.js
```diff
@@ -267,6 +343,12 @@ module.exports = async ({
     // Jobs
     errors = errors.concat(validateJobSchema(doc));
 
+    try {
+        verifyStages(doc.stages, doc.jobs);
+    } catch (error) {
+        errors.push(error);
+    }
```
sagar1312:
```suggestion
    errors = verifyStages(doc.stages, doc.jobs);
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
     return [];
 }
 
+/**
+ * Check there are no duplicate jobs in stages and all jobs listed exist
+ * @method verifyStages
+ * @param  {Object} stages Stages
+ * @param  {Object} jobs   Jobs
+ */
+function verifyStages(stages, jobs) {
```
sagar1312:
```suggestion
function validateStages(stages, jobs) {
```

lib/phase/functional.js
```diff
@@ -237,6 +258,61 @@ function validateWorkflowGraph(doc) {
     return [];
 }
 
+/**
+ * Check there are no duplicate jobs in stages and all jobs listed exist
+ * @method verifyStages
+ * @param  {Object} stages Stages
+ * @param  {Object} jobs   Jobs
```
sagar1312:
```suggestion
 * @param  {Object} stages Map of stage name to stage definition
 * @param  {Object} jobs     Map of job name to job configuration
```

lib/phase/functional.js
```diff
@@ -91,6 +92,26 @@ function validateBuildClusterAnnotation(doc, buildClusterFactory) {
...
 
+/**
+ * Ensure the triggers for a job are valid
+ * @method isJobTriggersValid
+ * @param  {Array}  triggers    List of requires configured for a job
+ * @param  {Array}  options     List of valid options to trigger a job
+ * @return {Boolean}
+ *
+ */
+function isJobTriggersValid(triggers, options) {
```
sagar1312, line 102: Let us keep it specific. For non-stage jobs we may never need such custom checks.
```suggestion
function isStageJobTriggersValid(jobTriggers, stageJobNames) {
```

test/index.test.js
```diff
@@ -241,6 +241,19 @@ describe('config parser', () => {
                             /YAMLException: Cannot have nonexistent job in stages: publish,deploy/
                         );
                     }));
+
+                it('returns an error if a job within a stage is triggered by jobs that are not within the same stage', () =>
+                    parser({
+                        yaml: loadData('pipeline-with-stages-and-invalid-triggers.yaml'),
```
sagar1312, line 254: Looks like `pipeline-with-stages-and-invalid-triggers.yaml` file got missed out during check-in

test/index.test.js
```diff
@@ -241,6 +241,19 @@ describe('config parser', () => {
                             /YAMLException: Cannot have nonexistent job in stages: publish,deploy/
                         );
                     }));
+
+                it('returns an error if a job within a stage is triggered by jobs that are not within the same stage', () =>
```
sagar1312, line 252: Can you also add a test for below scenario?

1. stage job does not have `requires`, but uses a job template which has `requires` having invalid trigger (Ex: `~commit`) for stage job.

sagar1312 (changes requested):

lib/phase/functional.js
VonnyJap: Validator does not produce an error for a normal workflow for a job trigger that does not exist. So maybe this is on purpose because sometimes trigger comes from an external pipeline?

Screenshot 2024-05-13 at 10 41 18 AM

test/index.test.js
VonnyJap: file is committed now

test/index.test.js
VonnyJap: So I added a test of job using a template that has `[~pr, ~commit]` as trigger, test is passing because apparently, the trigger is replaced with the setup job. So in the case of job [`boom`](https://github.com/screwdriver-cd/config-parser/blob/58df21d558d772777217ee919bb6441b66653f9c/test/data/pipeline-with-stages-and-invalid-triggers.yaml#L32-L33) that extends [this template](https://github.com/screwdriver-cd/config-parser/blob/58df21d558d772777217ee919bb6441b66653f9c/test/data/templateWithRequires.json#L1-L29), the final config is:
```
  boom: {
    image: 'node:4',
    steps: [ [Object], [Object] ],
    environment: {
      FOO: 'from template',
      BAR: 'foo',
      SD_TEMPLATE_FULLNAME: 'mytemplate',
      SD_TEMPLATE_NAME: 'mytemplate',
      SD_TEMPLATE_NAMESPACE: '',
      SD_TEMPLATE_VERSION: '1.2.3'
    },
    secrets: [ 'GIT_KEY' ],
    settings: { email: 'foo@example.com' },
    parameters: undefined,
    annotations: {},
    requires: [ '~stage@canary:setup' ],
    stage: { name: 'canary' },
    sourcePaths: [],
    templateId: 7754,
    commands: [ [Object], [Object] ]
  }
```
Is this by design or is this a bug?

VonnyJap: After checking, I think this is designed as such that the [stages is flattened](https://github.com/screwdriver-cd/config-parser/blob/5a0c19b985eef25d6e53cf1f2e308e44ce61486b/lib/phase/flatten.js#L686-L692) before the template flattened. So during the stages flattened, since there is no requires in `boom` job, it is immediately replaced it with setup job. And I think this is ok, user knows that when requires is omitted or it is an empty list `[]`, it means that it should be triggered by setup.

lib/phase/functional.js
```diff
@@ -6,6 +6,7 @@ const clone = require('clone');
 const WorkflowParser = require('screwdriver-workflow-parser');
 const SlackValidation = require('screwdriver-notifications-slack');
 const EmailValidation = require('screwdriver-notifications-email');
+// const YamlParser = require('js-yaml');
```
tkyi: commented code

VonnyJap: removed

lib/phase/functional.js
```diff
@@ -237,6 +257,66 @@ function validateWorkflowGraph(doc) {
...
+    const nonexistentJobsInStage = stageJobNames.filter(jobName => !jobNames.includes(jobName));
+
+    if (nonexistentJobsInStage.length > 0) {
+        errors.push(`Cannot have nonexistent job in stages: ${nonexistentJobsInStage}`);
+    }
+
+    Object.entries(stages).forEach(([stageName, stage]) => {
+        const jobsInStage = [...stage.jobs];
+        // Check if setup job is not in the list, then add it
+
```
tkyi, line 301:
```suggestion
```

lib/phase/functional.js
```diff
@@ -237,6 +257,66 @@ function validateWorkflowGraph(doc) {
...
+        if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
+            jobsInStage.push(`stage@${stageName}:setup`);
+        }
+        stage.jobs.forEach(jobName => {
+            if (jobs[jobName] && jobs[jobName].requires) {
+                if (!isJobTriggersValid(jobs[jobName].requires, jobsInStage)) {
+                    errors.push(
+                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers have to be jobs in ${stageName} stage.`
```
tkyi:
```suggestion
                        `${jobName} job has invalid requires: ${jobs[jobName].requires}. Triggers must be jobs from  ${stageName} stage.`
```

lib/phase/functional.js
```diff
@@ -237,6 +257,66 @@ function validateWorkflowGraph(doc) {
...
+    }
+
+    Object.entries(stages).forEach(([stageName, stage]) => {
+        const jobsInStage = [...stage.jobs];
+        // Check if setup job is not in the list, then add it
+
+        if (!jobsInStage.includes(`stage@${stageName}:setup`)) {
+            jobsInStage.push(`stage@${stageName}:setup`);
+        }
+        stage.jobs.forEach(jobName => {
```
tkyi, line 307:
```suggestion
        // Make sure requires jobs are all stage jobs
        stage.jobs.forEach(jobName => {
```

tkyi (approved):

PR #164, opened 2024-06-12
feat(3106): Allow some customized job fields for pipeline template

tkyi:
## Context

At present, users utilizing a pipeline template are restricted from making any modifications within the job configuration.
## Objective

This PR merges user job customizations for image, settings, environment, and requires within `job` field when using pipeline template. Also adds validation to make sure user job names match pipeline template job names.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/3106

## License

### DIFF
--- a/lib/phase/merge.js
+++ b/lib/phase/merge.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Hoek = require('@hapi/hoek');
+const clone = require('clone');
 const TEMPLATE_NAME_REGEX_WITH_NAMESPACE =
     require('screwdriver-data-schema').config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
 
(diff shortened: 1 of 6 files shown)


lib/phase/merge.js
```diff
@@ -70,6 +71,82 @@ function isEmpty(value) {
...
+            const oldSlackSettings = templateSettingsConfig.slack || {};
+            const newEmailSettings = yamlSettingsConfig.email || {};
+            const oldEmailSettings = templateSettingsConfig.email || {};
+
+            newJob.settings = {
+                ...templateSettingsConfig,
+                ...yamlSettingsConfig
+            };
+
+            if (!isEmpty(oldSlackSettings)) {
```
pritamstyz4ever, line 121: Do we need the not empty check, `oldSlackSettings` and `oldEmailSettings` are already set to {}, so expanding them `...{}` should be fine

pritamstyz4ever (approved):
Minor comment

lib/phase/merge.js
tkyi: I think it's still needed here. Otherwise we get an error `\"slack.0\" is not allowed`

PR #169, opened 2024-08-03
fix: add workflowgraph when parsing pipeline template

VonnyJap:
## Context

Workflow graph is not generated when parsing the pipeline template. 

## Objective

Generates the workflow graph when parsing the pipeline template. 

## License

### DIFF
--- a/index.js
+++ b/index.js
@@ -334,23 +334,10 @@ async function parsePipelineTemplate({ yaml }) {
 async function validatePipelineTemplate({ yaml, templateFactory }) {
     const pipelineTemplate = await parsePipelineTemplate({ yaml });
 
-    console.log('pipelineTemplate.config: ', pipelineTemplate.config);
-    console.log('templateFactory: ', templateFactory);
-
     // Merge template steps for validator
     const { newJobs } = await flattenTemplates(pipelineTemplate.config, templateFactory, true);
 
-    console.log('newJobs: ', newJobs);
-
     pipelineTemplate.config.jobs = newJobs;
-
-    // // Add workflowGraph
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -325,18 +333,24 @@ async function parsePipelineTemplate({ yaml }) {
...
+    console.log('newJobs: ', newJobs);
+
     pipelineTemplate.config.jobs = newJobs;
 
-    // Add workflowGraph
-    const { doc } = await phaseValidateFunctionality({
-        flattenedDoc: pipelineTemplate.config,
-        isPipelineTemplate: true
-    });
+    // // Add workflowGraph
```
tkyi: Commented code

index.js
```diff
@@ -325,18 +333,24 @@ async function parsePipelineTemplate({ yaml }) {
  */
 async function validatePipelineTemplate({ yaml, templateFactory }) {
     const pipelineTemplate = await parsePipelineTemplate({ yaml });
+
+    console.log('pipelineTemplate.config: ', pipelineTemplate.config);
```
tkyi: Console logs

tkyi (approved):

tkyi:
Screenshot 2024-08-11 at 3 49 49 PM
Might be related

VonnyJap:
I suspect that we somehow persist the workflow graph during the pipeline template creation so it failed to comply to this schema https://github.com/screwdriver-cd/data-schema/blob/master/config/base.js#L96-L109, the workaround for now is perhaps to add workflowGraph as one of the allowed keys, but I think by design, workflow graph should be calculated and not persisted.

PR #178, opened 2025-03-06
fix(3271): Enables sourcePaths to be used in stage [2]

yk634:
## Context

Enables the sourcePaths set in stage or the stage's setup job to function.

This change will not be available until the PR https://github.com/screwdriver-cd/data-schema/pull/591 is published.

## Objective

- If sourcePaths is set for both stage and setup job, the value of the setup job takes precedence.
- If sourcePaths is set to a normal job in stage, a validation error occurs.

## References

issue: https://github.com/screwdriver-cd/screwdriver/issues/3271

## License


yakanechi (approved):

y-oksaku (approved):

lib/phase/functional.js
```diff
@@ -315,6 +315,13 @@ function validateStages(stages, jobs) {
         });
     });
 
+    // If jobs in stages have sourcePaths set, throw error
+    stageJobNames.forEach(jobName => {
```
sagar1312, line 319: We should also perform the same check on stage teardown job to not allow `sourcePaths`

sagar1312 (changes requested):

lib/phase/functional.js
yk634: Created PR to modify to validate with data-schema.
https://github.com/screwdriver-cd/data-schema/pull/592

sagar1312 (approved):

PR #184, opened 2025-12-25
fix: Reduce the number of fetches for job templates

yakanechi:
## Context
Even when multiple jobs using the same template exist within a pipeline, each job attempts to retrieve the template from the database individually.

This poses a problem of increased load on the database for pipelines with a very large number of jobs.

## Objective

Reduce the number of fetches for same job templates

## License

### DIFF
--- a/lib/phase/flatten.js
+++ b/lib/phase/flatten.js
@@ -352,8 +352,8 @@ function handleMergeSharedStepsAnnotation({ sharedConfig, jobConfig, template })
  * Retrieve template and merge into job config
  *
  * @method mergeTemplateIntoJob
- * @param  {String}           jobName               Job name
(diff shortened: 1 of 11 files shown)


minghay (approved):

yk634:
In the following part, `template.config` is now reused across multiple jobs.
Because of that, directly referencing it may cause the underlying object to be mutated by subsequent `merge()` processing.

https://github.com/screwdriver-cd/config-parser/pull/184/files#diff-650a22c636f250a0b20166054e7ffaa7a337afe42b97be42acc2728c96107d0aR384

Would it make sense to deep-clone the template config per job, for example:

```diff
-        const newJob = template.config;
+        const newJob = clone(template.config);
```

yk634 (approved):

